PR #4, opened 2018-02-02
Adds column description

cmika:
This PR adds a column flag named `ColumnDescription`. This column flag is ignored during migrations.

### DIFF
--- a/src/Database/Orville/Internal/MigrateTable.hs
+++ b/src/Database/Orville/Internal/MigrateTable.hs
@@ -97,13 +97,14 @@ mkDropColumnDDL :: String -> Maybe SqlColDesc -> [String]
 mkDropColumnDDL _ Nothing = []
 mkDropColumnDDL name (Just _) = ["DROP COLUMN " ++ name]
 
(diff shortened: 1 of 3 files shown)


cmika:
I've updated this PR to restrict the version of Conduit to below 1.3

src/Database/Orville/Internal/MigrateTable.hs
```diff
@@ -122,7 +123,7 @@ mkFieldDDL :: FieldDefinition -> String
 mkFieldDDL (name, columnType, flags) =
         name ++ " " ++ sqlType ++ " " ++ flagSql
   where sqlType = mkTypeDDL columnType
-        flagSql = List.intercalate " " (notNull : map mkFlagDDL flags)
+        flagSql = List.intercalate " " (notNull : catMaybes (map mkFlagDDL flags))
```
qxjit: This should use `mapMaybe` rather than `catMaybes` and `map` separately.

qxjit:
@cmika You need to bump the orville version btw. I forgot to add that comment.

PR #6, opened 2018-04-04
Allows use of abitrary haskell types as primary keys

qxjit:
This requires adding a type parameter to TableDefinition to capture the
primary key type, as well as conversion functions related to it.

Note that this is probably not actually all the useful with Poppers
unless a similar change is made to FieldDefinition to track the Haskell
type associated with the fields. I'll add that as another commit here.

### DIFF
--- a/src/Database/Orville/Internal/Types.hs
+++ b/src/Database/Orville/Internal/Types.hs
@@ -24,6 +24,7 @@ import qualified Data.Time as Time
 
 import Database.Orville.Internal.Expr
 import Database.Orville.Internal.QueryKey
+import Database.Orville.Internal.SqlConversion
 
 type Record = Int
 
@@ -53,7 +54,7 @@ data ColumnFlag
   | Null
   | Unique
   | forall entity key. References (TableDefinition entity key)
-                                  FieldDefinition
+                                  (FieldDefinition key)
   | ColumnDescription String
 
 class ColumnDefault a where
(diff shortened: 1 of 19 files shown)


src/Database/Orville/Internal/Types.hs
```diff
@@ -181,33 +181,40 @@ getComponent getComp (ToSql serializer) =
...
+data TableDefinition entity key = TableDefinition
   { tableName :: String
       -- ^ The name of the table in the database.
   , tableFields :: [FieldDefinition]
       -- ^ A list of field definitions defining the table structure
   , tableSafeToDelete :: [String]
       -- ^ A list of any columns that may be deleted from the table by Orville.
       -- (Orville will never delete a column without being told it is safe)
-  , tableFromSql :: FromSql (entity Record)
+  , tableKeyToSql :: key -> SqlValue
```
jbrechtel: Nice. This makes for a good first step away from typeclasses.

qxjit:
@jbrechtel  Re the field definition change, it's much worse than that. *Every* `FieldDefinition` has to change, not just primary key ones.

src/Database/Orville/Internal/Where.hs
```diff
@@ -4,27 +4,42 @@ Copyright : Flipstone Technology Partners 2016-2018
...
+  to do at this moment.
+
+  Alternately, we could eliminate storing the entire FieldDefinition here,
+  thereby removing the need for ExistentialQuantification. Currently the
+  field definition is being used in the QueryKeyable instances for WhereCondition
+  for calls to qkOp an friends. Replacing FieldDefinition with just the field
+  name here would be nice. We would probably want a fully-fledged FieldName
+  type whech could provide the appropriate QueryKeyable instance. That then
+  raises questions about the ergonomics users creating FieldDefinition values
+  without requiring OverloadedStrings to be turned on.
```
telser: I wouldn't feel _too_ bad about requiring `OverloadedStrings` in light of the wins mentioned here.

telser:
This seems like a big win to me. Definitely going to require consumers to change some things, but that is expected for the 0.9 branch right?

qxjit:
@telser I share the sentiment that the items in the "good news" column outweigh temporary pain of the "bad news".

liminalisht:
This is looking really nice from my point of view.

PR #7, opened 2018-04-09
Makes a first pass at new FieldDefinition construction

qxjit:
This adds a set of helpers for creating FieldDefinitions with the
correct ColumnType values to match common Haskell types. This is more
step away from using convertible because using these predefined field
types together with `withConversion` will allow users to convert their
custom types to/from other normal Haskell types rather than needing to
define Convertible instances for their types with SqlValue.

You can see a very simple usage example in `Example/Schema/Virus.hs`,
though not complicated enough to exhibit all scenarios.

### DIFF
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -41,6 +41,20 @@ module Database.Orville.Core
   , Now(..)
   , ColumnType(..)
   , FieldDefinition
+  , textField
+  , fixedTextField
+  , dayField
+  , utcTimeField
+  , int32Field
+  , int64Field
+  , doubleField
+  , boolField
+  , automaticIdField
+  , searchVectorField
+  , nullableField
+  , withFlag
+  , withName
+  , withConversion
   , SomeField(..)
   , withPrefix
   , fieldName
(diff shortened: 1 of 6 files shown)


src/Database/Orville/Core.hs
```diff
@@ -41,6 +41,20 @@ module Database.Orville.Core
...
+  , int32Field
+  , int64Field
+  , doubleField
+  , boolField
+  , automaticIdField
+  , searchVectorField
+  , nullableField
+  , withFlag
+  , withName
+  , withConversion
```
liminalisht: I believe `nullableConversion` ought to be reexported by `Core` as well.

src/Database/Orville/Internal/SqlConversion.hs
```diff
@@ -24,6 +35,30 @@ data SqlConversion a = SqlConversion
...
+int32Conversion = sqlConvertible
+
+int64Conversion :: SqlConversion Int64
+int64Conversion = sqlConvertible
+
+doubleConversion :: SqlConversion Double
+doubleConversion = sqlConvertible
+
+boolConversion :: SqlConversion Bool
+boolConversion = sqlConvertible
```
liminalisht: love it

src/Database/Orville/Core.hs
qxjit: Good catch. I'll add it.

PR #8, opened 2018-04-11
Changes types to use app assigned primary key values

qxjit:
This essentially only changes the types involved in things so that
Orville doesn't apply the `entity` type variable to other variables.
Instead `entity` should be a fully applied concrete type.

Overall this makes the type signatures of functions simpler to read.
Occassionaly we have lost some clarity on how the `entity` and `key`
types are related. Only the presence of a `TableDefinition` in the
function signature enforces this relationship now.

We also lose any sense in the type system of "saved entities" vs
"unsaved entities". Where the type of `insertRecord` previously gave a
hint that entity given should not yet be saved now there is no such
hint.

This commit does not provide any facilities to help applications
generate approriate primary key values. More experimentation surrounding
that to follow.

### DIFF
--- a/src/Database/Orville/Internal/MigrateTable.hs
+++ b/src/Database/Orville/Internal/MigrateTable.hs
@@ -117,6 +117,7 @@ mkTypeDDL (VarText len) = "VARCHAR(" ++ show len ++ ")"
 mkTypeDDL (Date) = "DATE"
 mkTypeDDL (Timestamp) = "TIMESTAMP with time zone"
 mkTypeDDL TextSearchVector = "TSVECTOR"
+mkTypeDDL UUID = "UUID"
 
 mkFieldDDL :: FieldDefinition a -> String
 mkFieldDDL (name, columnType, flags, _) =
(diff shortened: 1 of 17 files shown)


test/Example/Data/Virus.hs
```diff
@@ -2,20 +2,35 @@ module Example.Data.Virus
...
+import Data.UUID (UUID)
+import Data.UUID.V4 as UUIDv4
 
 data Virus = Virus
   { virusId :: VirusId
   , virusName :: VirusName
   } deriving (Show, Eq)
 
+newVirus :: VirusName -> IO Virus
+newVirus name = liftA2 Virus newVirusId (pure name)
```
jbrechtel: Forgoing the applicative operator?

src/Database/Orville/Internal/MigrateTable.hs
```diff
@@ -139,8 +138,7 @@ mkCreateTableDDL tableDef =
     mkSomeFieldDDL (SomeField f) = mkFieldDDL f
 
 columnTypeSqlId :: ColumnType -> SqlTypeId
-columnTypeSqlId AutomaticId = SqlBigIntT
-columnTypeSqlId ForeignId = SqlBigIntT
+columnTypeSqlId UUID = SqlUnknownT "2950"
```
jbrechtel: I presume that SqlGUIDT isn't appropriate here? I know GUID != UUID generally but not even here?

qxjit: Although postgres has a`UUID` type natively, the type id we get back when describing the table is the one listed here, rather than `SqlGUIDT`. We have to match what postgres is returning for `migrateSchema` to know it doesn't need to do any migrations the second time around.

This is one (on many) reasons it would be nice for these sql interface types to be defined in a more extensible manner. We could provide the UUID type in a postgres-specific orville package.

qxjit:
I'm closing this at this point. I believe we've learned everything we can learn from this particular spike. We'll take this knowledge and roll in into another spike.

PR #10, opened 2018-04-27
Mmarabella/add project

mmarabella:
Add sample project to repo

### DIFF
--- a/sample-project/Example/Data/Student.hs
+++ b/sample-project/Example/Data/Student.hs
@@ -4,7 +4,9 @@ module Example.Data.Student
   , StudentName(..)
   ) where
 
-import Example.Data.Major ( Major(..), MajorId(..), MajorName(..), MajorCollege(..), majorCollegeConversion)
+import Data.Text (Text, pack, unpack)
+
+import Example.Data.Major ( Major(..), MajorId(..), MajorName(..), MajorCollege(..))
 
 data Student key = Student
   { studentId :: key
@@ -17,7 +19,7 @@ newtype StudentId = StudentId
   } deriving (Show, Eq, Ord)
 
 newtype StudentName = StudentName
-  { studentNameString :: String
+  { studentNameText :: Text
   } deriving (Show, Eq)
 
 
--- a/sample-project/Example/Data/Student.hs
+++ b/sample-project/Example/Data/Student.hs
@@ -2,25 +2,23 @@ module Example.Data.Student
   ( Student(..)
   , StudentId(..)
   , StudentName(..)
-  , StudentMajor(..)
   ) where
+
+import Example.Data.Major ( Major(..), MajorId(..), MajorName(..), MajorCollege(..), majorCollegeConversion)
 
 data Student key = Student
   { studentId :: key
   , studentName :: StudentName
-  , studentMajor :: StudentMajor
+  , studentMajor :: MajorId
   } deriving (Show, Eq)
 
 newtype StudentId = StudentId
-  { getStudentId :: String
+  { studentIdInt :: Int
   } deriving (Show, Eq, Ord)
 
 newtype StudentName = StudentName
-  { getStudentName :: String
+  { studentNameString :: String
   } deriving (Show, Eq)
 
-newtype StudentMajor = StudentMajor
-  { getStudentMajor :: String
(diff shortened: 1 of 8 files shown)


sample-project/Main.hs
```diff
@@ -0,0 +1,147 @@
...
+import Example.Data.Student ( Student(..), StudentId(..), StudentName(..), StudentMajor(..))
+import Example.Schema.Student ( studentIdField, studentNameField, studentMajorField )
+import Example.SchemaStudent (studentSchema, studentTable)
+
+
+main :: IO ()
+main = do
+  putStrLn "Connecting to db"
+  -- To do: reference environment variable
+  let connStr = "host=testdb user=orville_test"
```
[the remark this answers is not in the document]
telser: This. In particular because I completely plan on continuing to develop on FreeBSD :D

sample-project/Example/Data/Student.hs
```diff
@@ -0,0 +1,26 @@
...
+  ( Student(..)
+  , StudentId(..)
+  , StudentName(..)
+  , StudentMajor(..)
+  ) where
+
+data Student key = Student
+  { studentId :: key
+  , studentName :: StudentName
+  , studentMajor :: StudentMajor
```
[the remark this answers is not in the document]
mmarabella: @qxjit I think I have fixed all the issues except this one. I thought I understood how to do it, but since I've been running into issues for the last few days, I wanted to make sure I'm on the right track. Should I just be using the ForeignId ColumnType?

sample-project/Example/Schema/Student.hs
```diff
@@ -0,0 +1,91 @@
...
+  ( "name"
+  , O.VarText 255
+  , []
+  , O.sqlConversionVia studentNameString StudentName O.sqlConvertible)
+
+studentMajorField :: O.FieldDefinition MajorId
+studentMajorField =
+  ( "major"
+  , O.ForeignId
+  , []
```
jbrechtel: You want a `References` `ColumnFlag` that points to the `majors` table and the `majorIdField` so that Orville will create the foreign key constraint when creating the table.

mmarabella: @jbrechtel Thank you! That makes sense. I'll fix that and add an example that violates the constraint.

qxjit: Actually there's a `FieldDefinition` helper to set this up correctly. I'd like for *all* the fields here to use the field definition helpers, not just `automaticIdField`

sample-project/Example/Schema/Student.hs
```diff
@@ -0,0 +1,91 @@
...
+    , O.tblSetKey = \key entity -> entity {studentId = key}
+    , O.tblSafeToDelete = []
+    , O.tblComments = O.noComments
+    }
+
+studentIdField :: O.FieldDefinition StudentId
+studentIdField =
+  O.withConversion 
+  (O.withFlag (O.automaticIdField "id") O.PrimaryKey) 
+  (O.sqlConversionVia studentIdInt StudentId)
```
qxjit: The `with*` were designed to be used with backtick-style infix notation, like in the `Schema.Virus` module for the tests.

sample-project/Example/Data/Student.hs
```diff
@@ -0,0 +1,24 @@
...
+  , studentName :: StudentName
+  , studentMajor :: MajorId
+  } deriving (Show, Eq)
+
+newtype StudentId = StudentId
+  { studentIdInt :: Int
+  } deriving (Show, Eq, Ord)
+
+newtype StudentName = StudentName
+  { studentNameString :: String
```
qxjit: We should by using `Text` rather than `String` for all these types.

sample-project/Example/Schema/Student.hs
mmarabella: @qxjit core did not import the foreign key FieldDefinition helper, so I added it to core's import list. Would it be better if I imported foreignKeyField directly from the FieldDefinition module? Otherwise, I think I fixed all the field definitions.

qxjit:
@mmarabella Let's merge this to the `0.9_development` branch now. I'm sure there is more we'll want to extend it more in the future, but this is a nice first pass.

PR #11, opened 2018-04-27
[ch12060] Separates saved/unsaved types in TableDefinition

qxjit:
TableDefinition previously accept a type constructor with one parameter
and constructed the saved/unsaved versions of the type by applied the
appropriate key type. This made the TableDefinition read nicely, but
forced the types used with it to be more sophisticated.

This separates the one parameter into two parameters of types (not type
constructors) that must be explicitly passed. The notable effects of
this are:

1) The abolishement of the key as a type parameter, enabling it to be
   specified as an application provided value (yet to be tested)

2) A bit of simplification of the return types of Orville functions.
   They now return the `fullEntity` type instead of `entity key`.

3) A bit of complication of the type signature of Orville functions
   involved TableDefinition because it now has two type params.

4) It still supports defining the record types in the old way simply
   by filling in the correct types to the TableDefinition. See the
   Virus example.

I have done nothing here to attempt to rectify the RelationMap to work
with the new scheme. I will do that before I consider this spike
complete. I will also add some other schema examples exploring
alternative usages patterns of records construction to see what works
and what doesn't, including application assigned primary keys.


jbrechtel:
This has an as-of-yet unpushed change to make it 3 parameters (C-R-U), right?

PR #13, opened 2018-06-18
Add ability to use qualified fields in WHERE

exodrifter:
Add ability to specify qualified table names in `WHERE` clauses. `String` was used instead of `TableDefinition` since `TableDefinition` requires a second type parameter.

A longer-term solution should be pursued; this is just a quick fix for now.

### DIFF
--- a/src/Database/Orville/Internal/Where.hs
+++ b/src/Database/Orville/Internal/Where.hs
@@ -24,6 +24,7 @@ data WhereCondition =
   | Or [WhereCondition]
   | And [WhereCondition]
   | AlwaysFalse
+  | Qualified String WhereCondition
 
 instance QueryKeyable WhereCondition where
   queryKey (BinOp op field value) = qkOp2 op field value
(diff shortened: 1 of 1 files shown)


src/Database/Orville/Internal/Where.hs
```diff
@@ -98,6 +100,9 @@ whereConditionSql (And conds) = List.intercalate " AND " condsSql
   where condsSql = map condSql conds
         condSql c = "(" ++ whereConditionSql c ++ ")"
 
+whereConditionSql (Qualified tableNameStr cond) =
+  tableNameStr ++ "." ++ whereConditionSql cond
```
qxjit: Instead of hard coding the `"."` behavior here, this should pass the table name down recursively to let each node in the clause tree use the qualification if it needs to.

This probably involves renaming `whereConditionSql` to a new internal function and adding a `Maybe String` parameter to carry the optional qualification through the recursion. Then you can add a replacement `whereConditionSql` that begins the recursion with `Nothing` as the default qualification string.

src/Database/Orville/Internal/Where.hs
```diff
@@ -24,6 +24,7 @@ data WhereCondition =
   | Or [WhereCondition]
   | And [WhereCondition]
   | AlwaysFalse
+  | Qualified String WhereCondition
```
qxjit: You could use a `TableDefinition` here if you use `ExistentialQuantification`. This is how `SchemaItem` accomplishes this task today.

exodrifter:
Need to change the PR to use `0.9_development` instead of `master`.

PR #14, opened 2018-06-20
Add ability to generate where SQL clauses with qualified field names

exodrifter:
Add the ability to get the generated `where` SQL clauses for qualified field names.

Right now, Orville lacks a lot of features. However, it is still desirable to replace some hard-coded SQL strings with automatically-generated ones. This allows a user to slowly replace SQL strings in the application as Orville becomes more feature complete.

### DIFF
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -98,6 +98,7 @@ module Database.Orville.Core
   , whereOr
   , whereIn
   , whereNotIn
+  , whereQualified
   , isNull
   , isNotNull
   , (.==)
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -101,6 +101,7 @@ module Database.Orville.Core
   , whereQualified
   , isNull
   , isNotNull
+  , whereConditionSql
   , (.==)
   , (.<>)
   , (.<-)
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -101,7 +101,6 @@ module Database.Orville.Core
   , whereQualified
   , isNull
   , isNotNull
-  , whereConditionSql
   , (.==)
   , (.<>)
   , (.<-)
--- a/src/Database/Orville/Internal/Where.hs
+++ b/src/Database/Orville/Internal/Where.hs
@@ -6,7 +6,27 @@ License   : MIT
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ExistentialQuantification #-}
 
-module Database.Orville.Internal.Where where
+module Database.Orville.Internal.Where
+( WhereCondition
+, (.==)
+, (.<>)
+, (.>)
+, (.>=)
+, (.<)
+, (.<=)
+, (.<-)
+, (%==)
+, whereConditionSql
+, whereConditionValues
+, whereAnd
+, whereOr
+, whereIn
+, whereNotIn
+, isNull
+, isNotNull
+, whereClause
+, whereValues
+) where
 
 import qualified Data.List as List
 import Database.HDBC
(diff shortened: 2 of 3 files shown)


src/Database/Orville/Internal/Where.hs
```diff
@@ -80,27 +103,39 @@ fieldDef .<- as = In fieldDef (List.nub $ map (fieldToSqlValue fieldDef) as)
...
+    condSql c = "(" ++ internalWhereConditionSql tableDef c ++ ")"
+internalWhereConditionSql tableDef (And conds) = List.intercalate " AND " condsSql
   where
     condsSql = map condSql conds
-    condSql c = "(" ++ whereConditionSql c ++ ")"
+    condSql c = "(" ++ internalWhereConditionSql tableDef c ++ ")"
+internalWhereConditionSql _ (Qualified tableDef cond) = internalWhereConditionSql (Just tableDef) cond
+
+getQualifiedFieldName :: Maybe (TableDefinition a b c) -> FieldDefinition d -> String
+getQualifiedFieldName maybeTableDef fieldDef =
```
qxjit: This should just be named `qualifiedFieldName`.

src/Database/Orville/Core.hs
```diff
@@ -98,8 +98,10 @@ module Database.Orville.Core
   , whereOr
   , whereIn
   , whereNotIn
+  , whereQualified
   , isNull
   , isNotNull
+  , whereConditionSql
```
qxjit: I don't know that I'm prepared for us to expose `whereConditionSql` an a part of the public api. We should probably discuss with @liminalisht and @cmika about how they intend to use this.

test/QualifiedTest.hs
```diff
@@ -0,0 +1,169 @@
...
+      [ testCase "Qualified Names" $ do
+          run (TestDB.reset schema)
+          void $ run (O.insertRecord orderTable foobarOrder)
+          void $ run (O.insertRecord orderTable badOrder)
+          void $ run (O.insertRecord customerTable aliceCustomer)
+          void $ run (O.insertRecord customerTable bobCustomer)
+
+          let opts = O.where_
+                   $ O.whereQualified customerTable
+                   $ (customerNameField .== (CustomerName "Alice"))
```
qxjit: This could use `customerName aliceCustomer` instead of reconstructing `CustomerName "Alice"`

test/QualifiedTest.hs
```diff
@@ -0,0 +1,169 @@
...
+          result <- run (S.runSelect $ completeOrderSelect opts)
+
+          case length result of
+            0 -> assertFailure "Expected CompleteOrder, but no records returned"
+            1 -> assertEqual
+                "Order returned didn't match expected result"
+                "Alice"
+                (customer $ result !! 0)
```
qxjit: We should avoid `!!` in general, because it is a partial function. Even though it's clearly safe in this instead, it's better to avoid using partial functions at all and restructure code to not require them instead. Doing so help avoid bugs creeping in because the preconditions that ensure the partial function succeeds are changed by future changes in the code.

test/QualifiedTest.hs
```diff
@@ -0,0 +1,169 @@
...
+            _ -> assertFailure "Expected one record, but got many records"
```
qxjit: This hides the information of what multiple records were returned. You could get around this by simply constructing the `CompleteOrder` results you expect and using `assertEqual` with the expected results with the actual ones.

test/QualifiedTest.hs
```diff
@@ -0,0 +1,169 @@
...
+
+foobarOrder :: Order
+foobarOrder = Order
+  { orderId = OrderId 1
+  , customerFkId = CustomerId 1
+  , orderName = OrderName "foobar"
+  }
+
+badOrder :: Order
+badOrder = Order
```
qxjit: rather than simple `badOrder` here, a better name would make it more clear what case you're trying to set up. Maybe `orderWithNameAlice` or `orderWithNameThatOverlapsCustomer`? Naming things is hard 🙁

src/Database/Orville/Internal/Where.hs
```diff
@@ -6,7 +6,27 @@ License   : MIT
...
+, whereAnd
+, whereOr
+, whereIn
+, whereNotIn
+, whereQualified
+, isNull
+, isNotNull
+, whereClause
+, whereValues
+) where
```
qxjit: ❤️  explicit exports

PR #15, opened 2018-06-22
Add whereLike

exodrifter:
Add support for `LIKE` queries.

`whereLike` expects the field to be compared against and the raw sql string to compare with. This allows users to put `%` or `_` anywhere they please. It also automatically calls `LOWER` for both arguments to `LIKE`, since I expect that to be desirable.

### DIFF
--- a/src/Database/Orville/Internal/Where.hs
+++ b/src/Database/Orville/Internal/Where.hs
@@ -170,7 +170,7 @@ whereIn :: FieldDefinition a -> [a] -> WhereCondition
 whereIn fieldDef values = In fieldDef (map (fieldToSqlValue fieldDef) values)
 
 whereLike :: FieldDefinition a -> String -> WhereCondition
-whereLike fieldDef raw = Like fieldDef (SqlString raw)
+whereLike fieldDef raw = Like fieldDef (toSql raw)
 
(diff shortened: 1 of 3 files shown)


src/Database/Orville/Internal/Where.hs
```diff
@@ -117,6 +121,8 @@ internalWhereConditionSql tableDef (In fieldDef values) =
   qualifiedFieldName tableDef fieldDef ++ " IN (" ++ quesses ++ ")"
   where
     quesses = List.intercalate "," (map (const "?") values)
+internalWhereConditionSql tableDef (Like fieldDef raw) =
+  "LOWER(" ++ (qualifiedFieldName tableDef fieldDef) ++ ") LIKE LOWER('" ++ raw ++ "')"
```
qxjit: We should not be doing raw sql interpolation on the value here. This should be implemented as a placeholder and the value to compare against should be put in the list of values from `whereConditionValues`

Raw interpolation of values that might very will be end user provided is a major security risk. If anyone implemented a backend using this `Like` to handle fields typed into a search field by a user, that user would have raw sql power.

Re: `LOWER`, I don't particular like encoding this preference into Orville itself. It makes Orville's `Like` not semantically equivalent to the database's definition of `LIKE`.

qxjit:
❤️

PR #17, opened 2018-06-26
Add support for whereLikeInsensitive

exodrifter:
Adds support for `ILIKE` queries.

Also updated the qualified where test cases to have a separate test case for `LIKE`, `ILIKE`, and `==`.

### DIFF
--- a/test/QualifiedTest.hs
+++ b/test/QualifiedTest.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 
 module QualifiedTest where
 
@@ -9,21 +10,18 @@ import qualified TestDB as TestDB
 
 import Control.Monad (void)
 import Data.Int (Int64)
+import Database.Orville ((.==))
 import Database.Orville.Expr (aliased, qualified)
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.HUnit (assertEqual, testCase)
 
-test_qualified_name :: TestTree
-test_qualified_name =
+test_where_qualified :: TestTree
+test_where_qualified =
   TestDB.withOrvilleRun $ \run ->
     testGroup
-      "QualifiedTest"
-      [ testCase "Qualified Names" $ do
-          run (TestDB.reset schema)
-          void $ run (O.insertRecord orderTable foobarOrder)
-          void $ run (O.insertRecord orderTable orderNamedAlice)
-          void $ run (O.insertRecord customerTable aliceCustomer)
-          void $ run (O.insertRecord customerTable bobCustomer)
+      "Qualified where query"
+      [ testCase "Qualified where like" $ do
+          resetDB run
           let opts =
                 O.where_ $
                 O.whereQualified customerTable $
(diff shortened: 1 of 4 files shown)


test/QualifiedTest.hs
```diff
@@ -33,8 +31,38 @@ test_qualified_name =
             "Order returned didn't match expected result"
             [CompleteOrder {order = "foobar", customer = "Alice"}]
             result
+      , testCase "Qualified where like insensitive" $ do
```
qxjit: It's starting to look like this should be 3 separate tests -- 1 about qualified, 1 about like, and 1 about like insensitive. That would also indicate that probably this file is more like a `WhereConditionTest` module rather than `QualifiedTest` module.

test/QualifiedTest.hs
```diff
@@ -33,8 +31,38 @@ test_qualified_name =
...
+            result
       ]
 
+resetDB :: (forall a. TestDB.TestMonad a -> IO a) -> IO ()
+resetDB run = do
+  run (TestDB.reset schema)
+  void $ run (O.insertRecord orderTable foobarOrder)
+  void $ run (O.insertRecord orderTable orderNamedAlice)
+  void $ run (O.insertRecord customerTable aliceCustomer)
+  void $ run (O.insertRecord customerTable bobCustomer)
```
qxjit: The one thing I don't like about extracting the inserting of the records out into this function is that it makes the tests harder to read individually. The data in the database for each test is really important for understanding the assertion being made in the test.

test/WhereConditionTest.hs
```diff
@@ -33,6 +34,36 @@ test_qualified_name =
             "Order returned didn't match expected result"
             [CompleteOrder {order = "foobar", customer = "Alice"}]
             result
+      , testCase "Qualified where like insensitive" $ do
+          run (TestDB.reset schema)
+          void $ run (O.insertRecord orderTable foobarOrder)
+          void $ run (O.insertRecord orderTable orderNamedAlice)
+          void $ run (O.insertRecord customerTable aliceCustomer)
+          void $ run (O.insertRecord customerTable bobCustomer)
```
qxjit: The tests for `whereLike` and `whereLikeInsensitive` don't need to be using `whereQualified` at all to test what they need to test. It would be nicer to have these tests only need to know about a single table instead needing them to also be concerned about the "qualified" feature that is covered by the test below.

exodrifter: Oh, duh. My bad.

test/WhereConditionTest.hs
```diff
@@ -105,6 +130,16 @@ newtype OrderName = OrderName
   { unOrderName :: T.Text
   } deriving (Show, Eq)
 
+orderSelect :: O.SelectOptions -> S.Select Order
+orderSelect = S.selectQuery buildOrder (S.fromClauseTable orderTable)
+
+buildOrder :: O.FromSql Order
+buildOrder =
```
qxjit: You don't have to add this. The `TableDefinition` already contains everything you need to do this. If you still want to use `S.Select` then you can use `selectQueryTable` to build the select. Alternatively, you can just use `selectAll` from core.

PR #18, opened 2018-07-02
Add support for SELECT DISTINCT

exodrifter:
Adds support for `SELECT DISTINCT` queries. A distinct query can be done by using `whereDistinct` or by modifying `SelectOptions`.

### DIFF
--- a/sample-project/Main.hs
+++ b/sample-project/Main.hs
@@ -156,7 +156,7 @@ selectFirstTest = do
 
 selectAllTest :: O.OrvilleT Postgres.Connection IO [Student StudentId]
 selectAllTest = do
-  let options = O.SelectOptions mempty mempty mempty mempty mempty
+  let options = O.SelectOptions mempty mempty mempty mempty mempty mempty
   O.selectAll studentTable options
 
 findRecordTest :: O.OrvilleT Postgres.Connection IO (Maybe (Student StudentId))
--- a/src/Database/Orville/Internal/SelectOptions.hs
+++ b/src/Database/Orville/Internal/SelectOptions.hs
@@ -112,11 +112,11 @@ selectOffsetClause opts =
     Nothing -> ""
     Just _ -> "OFFSET ?"
 
+distinct :: SelectOptions
+distinct = SelectOptions (First $ Just True) mempty mempty mempty mempty mempty
+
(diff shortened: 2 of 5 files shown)


src/Database/Orville/Internal/SelectOptions.hs
```diff
@@ -105,18 +113,21 @@ selectOffsetClause opts =
     Just _ -> "OFFSET ?"
 
 where_ :: WhereCondition -> SelectOptions
-where_ clause = SelectOptions [clause] mempty mempty mempty mempty
+where_ clause = SelectOptions mempty [clause] mempty mempty mempty mempty
+
+whereDistinct :: WhereCondition -> SelectOptions
+whereDistinct clause = SelectOptions (First $ Just True) [clause] mempty mempty mempty mempty
```
qxjit: No need to involve `where` here, this should just be `distinct :: SelectOptions`

test/WhereConditionTest.hs
```diff
@@ -19,7 +19,19 @@ test_where_condition =
...
     testGroup
       "Where condition queries"
-      [ testCase "Where like" $ do
+      [ testCase "Where distinct" $ do
+          run (TestDB.reset schema)
+          void $ run (O.insertRecord orderTable foobarOrder)
+          void $ run (O.insertRecord orderTable anotherFoobarOrder)
+          let opts =
+                O.whereDistinct $
+                (orderNameField .== orderName foobarOrder)
```
qxjit: This condition isn't necessary for the test. You could just get all the records without any `where`

sample-project/Main.hs
```diff
@@ -156,7 +156,7 @@ selectFirstTest = do
 
 selectAllTest :: O.OrvilleT Postgres.Connection IO [Student StudentId]
 selectAllTest = do
-  let options = O.SelectOptions mempty mempty mempty mempty mempty
+  let options = O.SelectOptions mempty mempty mempty mempty mempty mempty
```
qxjit: This makes me realize we probably should not be exporting the `SelectOptions` constructor in the future. That should really be internal. But no need to fix that for the sake of this PR.

exodrifter: I was thinking while doing this that there really ought to be an "empty select options" function. It could be used in a lot of places internally as well.

PR #19, opened 2018-07-06
liminalisht/qualifiedtablefromsql

liminalisht:
This PR adds a function to return a `FromSql fullEntity` given a `TableDefinition fullEntity writeEntity key` in such a way that the selections are qualified by the table's name.

An example use case is where we want to join on tables that have column names that conflict. We'd like (in such a case) to reuse the fromsql derived from the table, but we can't b/c we get an ambilguous reference error unless qualified.
```
"SqlError {seState = \"42702\", seNativeError = 7, seErrorMsg = \"execute: PGRES_FATAL_ERROR: ERROR:  column reference \\\"id\\\" is ambiguous\\nLINE 1: SELECT \\\"id\\\", \\\"supplier_order_id\\\"..."
```

With this function, we can reuse that `FromSql` qualified by the table's name as so:
```
f = runSelect $ selectQuery fromSql fromClause selectOptions
  where
    fromSql    = qualifiedTableFromSql someTable
    fromClause = fromClauseRaw $
      " FROM some_table \
      \ LEFT JOIN some_other_table   \
      \ ON some_table.supplier_order_id = some_other_table.supplier_order_id "
    selectOptions   = where_ whereConditions
    whereConditions = whereAnd
      [ whereQualified someTable $ ...
      , whereQualified someOtherTable $ ...
      ]
```

### DIFF
--- a/src/Database/Orville/Internal/RelationalMap.hs
+++ b/src/Database/Orville/Internal/RelationalMap.hs
@@ -26,6 +26,7 @@ import Control.Monad (join, when)
 import Control.Monad.Reader (ask)
 import Control.Monad.State (modify)
 
+import Database.Orville.Internal.Expr
 import Database.Orville.Internal.FieldDefinition
 import Database.Orville.Internal.FromSql
 import Database.Orville.Internal.SqlConversion
@@ -86,7 +87,7 @@ mkTableDefinition ::
 mkTableDefinition (TableParams {..}) =
   TableDefinition
     { tableFields = fields tblMapper
-    , tableFromSql = mkFromSql tblMapper
+    , tableFromSql = mkQualifiedFromSql tblName tblMapper
     , tableToSql = mkToSql tblMapper
     , tablePrimaryKey = tblPrimaryKey
     , tableName = tblName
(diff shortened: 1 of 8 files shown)


qxjit:
@liminalisht If we want to make this the default behavior for all queries, probably the easiest place to add the qualified is in `mkFromSql` in the `RelationalMap` module. At the moment than `RM_Field` gets turned into a `FromSql`, you could replace the `col field` with something that produces a qualified select. You'll need to add the table name to the `mkFromSql` signature

liminalisht:
Instead of modifying `mkFromSql`, I've gone the route of having `mkTableDefinition` invoke a `mkQualifiedFromSql` function that itself invokes `mkFromSql`. Orville tests pass, and tests on our end pass too.

qxjit:
@liminalisht Did the other approach not work? Doing it this way seems rather round-about

liminalisht:
Ok please take another look at you're convenience. I think this is the way you wanted it.

src/Database/Orville/Internal/RelationalMap.hs
```diff
@@ -176,6 +177,12 @@ fields (RM_ReadOnly rm) =
   where
     someFieldWithFlag flag (SomeField f) = SomeField (f `withFlag` flag)
 
+mkQualifiedFromSql :: String -> RelationalMap a b -> FromSql b
+mkQualifiedFromSql tableName relationalMap = unqualifiedFromSql {fromSqlSelects = qualifiedSelects}
+  where
+    unqualifiedFromSql = mkFromSql relationalMap
+    qualifiedSelects = fmap (`qualified` (NameForm tableName)) $ fromSqlSelects unqualifiedFromSql
+
```
qxjit: I don't like this approach of modifying the `fromSqlSelects` after having built the `FromSql`. We should invest in making the tools to build the correct select name in the first place, rather than building the wrong one and then modifying the data structure after the fact.

liminalisht:
I've got orville tests passing but HEB tests still fail in a few cases. There are still a few places in orville that I'm missing where we should be qualifying by table name when supplying the builder (result set).

I'm going to close this for now and hopefully revisit it at some later point.

PR #20, opened 2018-07-09
Add support for making raw WHERE conditions

exodrifter:
The idea behind this PR is to leverage the current capability of Orville to generate most `WHERE` conditions, but still allow custom-written where clauses when needed, such as those that contain subqueries.

### DIFF
--- a/src/Database/Orville/Internal/Where.hs
+++ b/src/Database/Orville/Internal/Where.hs
@@ -24,6 +24,7 @@ module Database.Orville.Internal.Where
   , whereLikeInsensitive
   , whereNotIn
   , whereQualified
+  , whereRaw
   , isNull
   , isNotNull
   , whereClause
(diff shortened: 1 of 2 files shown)


src/Database/Orville/Internal/Where.hs
```diff
@@ -71,6 +72,7 @@ data WhereCondition
   | AlwaysFalse
   | forall a b c. Qualified (TableDefinition a b c)
                             WhereCondition
+  | Raw String
```
qxjit: I think that instead of simply adding `Raw` here, it's worth stopping for a moment and figuring out how `WhereCondition` and the whole `Expr` module should be unified. `Expr` already allows for raw sql to be provided at any given part of an expression, so fitting `WhereCondition` into that framework would automatically support providing raw where conditions.

exodrifter: That sounds like a much better idea.

exodrifter:
See #21 for continued discussion on this PR.

PR #21, opened 2018-07-16
Add ability to qualify any NameForm

exodrifter:
The ability to qualify any `NameForm` was added to make integrating the `Expr` module and the `Where` module easier.

Since `NameForm` is a way to refer to any name in a SQL query, we want to be able to qualify the name if needed. In order to do this, I added a new field to `NameForm` and moved `qualified` to `NameExpr`.

I left the functionality of `instance IsString NameForm where` intact since generally speaking, the user will want to start with an unqualified name and then use `qualified` as needed.

### DIFF
--- a/src/Database/Orville/Internal/Expr/Expr.hs
+++ b/src/Database/Orville/Internal/Expr/Expr.hs
@@ -34,6 +34,16 @@ instance IsString RawExpr where
 newtype Expr a =
   Expr (Either RawExpr a)
 
+class QualifySql form where
+  qualified :: form -> String -> form
+
+instance QualifySql RawExpr where
+  qualified expr _ = expr
+
+instance QualifySql a => QualifySql (Expr a) where
+  qualified a table = qualified a table
+  qualified raw _ = raw
+
 class GenerateSql expr where
   generateSql :: expr -> RawExpr
 
--- a/src/Database/Orville/Internal/Expr/Expr.hs
+++ b/src/Database/Orville/Internal/Expr/Expr.hs
@@ -37,12 +37,9 @@ newtype Expr a =
 class QualifySql form where
   qualified :: form -> String -> form
 
-instance QualifySql RawExpr where
-  qualified expr _ = expr
-
 instance QualifySql a => QualifySql (Expr a) where
-  qualified a table = qualified a table
-  qualified raw _ = raw
+  qualified (Expr (Right a)) table = Expr . Right $ qualified a table
+  qualified (Expr (Left raw)) _ = Expr . Left $ raw
 
 class GenerateSql expr where
   generateSql :: expr -> RawExpr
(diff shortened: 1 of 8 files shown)


test/WhereConditionTest.hs
```diff
@@ -80,12 +81,12 @@ completeOrderSelect = S.selectQuery buildCompleteOrder orderCustomerFrom
...
-    (S.selectField orderNameField `qualified` "order" `aliased` "order_name") <*>
-  O.col
-    (S.selectField customerNameField `qualified` "customer" `aliased`
-     "customer_name")
+    CompleteOrder <$>
+    O.col (E.selectColumn orderNameForm `aliased` "order_name") <*>
+    O.col (E.selectColumn customerNameForm `aliased` "customer_name")
+  where
+    orderNameForm = (O.fieldToNameForm orderNameField) `qualified` "order"
+    customerNameForm = (O.fieldToNameForm customerNameField) `qualified` "customer"
```
qxjit: I don't think I'm prepared to give up the ergonomics of `selectField` in favor of requiring the user to use `selectColumn` and `fieldToNameForm` in conjunction for this use case.

Maybe we should consider making `qualified` a method in a typeclass so that it can be used with both `NameForm` and `SelectForm`. The `SelectForm` instance would delegate to the `NameForm` instance to qualify the `NameForm` it contains.

exodrifter: I really like that idea

src/Database/Orville/Internal/Expr/Expr.hs
```diff
@@ -34,6 +34,16 @@ instance IsString RawExpr where
 newtype Expr a =
   Expr (Either RawExpr a)
 
+class QualifySql form where
+  qualified :: form -> String -> form
+
+instance QualifySql RawExpr where
+  qualified expr _ = expr
```
qxjit: Does this instance make sense? If it simply ignores the qualification, why bother providing an instance at all?

src/Database/Orville/Internal/Expr/Expr.hs
```diff
@@ -34,6 +34,16 @@ instance IsString RawExpr where
...
+
+instance QualifySql a => QualifySql (Expr a) where
+  qualified a table = qualified a table
+  qualified raw _ = raw
```
qxjit: I like this instance ergonomically to allow `qualified` to be used with any `Expr` that corresponds to a qualifiable `Form`. The implementation is not correct, however. I think you intended to unpack the `Left` and `Right` constructors inside the `Expr`, but you didn't actually do that. The first branch of `qualified` will result in an infinite loop here (because it just calls itself) while the second branch will never get hit.

exodrifter: Oh, I see. I did try unpacking `Left` and `Right`, but the compiler kept complaining about mismatched types for some reason. I guess I resolved it incorrectly, I'll take another look at it.

src/Database/Orville/Internal/Expr/Expr.hs
exodrifter: I thought it would be necessary for `instance QualifySql a => QualifySql (Expr a)` at first, but that's not true. I'll remove it.

PR #22, opened 2018-07-18
Initial WhereExpr

exodrifter:
Adds the initial `WhereExpr` type.

This PR is one part of a potential series of PRs to slowly replace the functionality in `WhereCondition` with `WhereExpr`.

This PR only implements the binary operators. If this PR is approved, then I will do the same incrementally to the rest of the `WhereCondition` sumtypes.

The original API is preserved in this PR. It should be noted that if someone qualifies a `WhereForm` and then uses `Qualified` on the resulting `WhereCondition`, the `WhereCondition` qualification will take precedence over the `WhereForm` one. This is done in favor of making sure the original API still works until it is completely replaced by `WhereExpr`.

### DIFF
--- a/src/Database/Orville/Internal/Expr/NameExpr.hs
+++ b/src/Database/Orville/Internal/Expr/NameExpr.hs
@@ -11,6 +11,7 @@ import Data.Monoid
 import Data.String
 
 import Database.Orville.Internal.Expr.Expr
+import Database.Orville.Internal.QueryKey
 
 type NameExpr = Expr NameForm
 
@@ -30,6 +31,9 @@ instance IsString NameForm where
 
 instance QualifySql NameForm where
   qualified form table = form {nameFormTable = Just table}
+
+instance QueryKeyable NameForm where
+  queryKey = QKField . nameFormName
 
 instance GenerateSql NameForm where
   generateSql (NameForm Nothing name) =
--- a/src/Database/Orville/Internal/Expr/NameExpr.hs
+++ b/src/Database/Orville/Internal/Expr/NameExpr.hs
@@ -33,7 +33,7 @@ instance QualifySql NameForm where
   qualified form table = form {nameFormTable = Just table}
 
 instance QueryKeyable NameForm where
-  queryKey = QKField . nameFormName
+  queryKey = QKField . unescapedName
 
 instance GenerateSql NameForm where
   generateSql (NameForm Nothing name) =
--- a/src/Database/Orville/Internal/Expr/WhereExpr.hs
+++ b/src/Database/Orville/Internal/Expr/WhereExpr.hs
@@ -65,7 +65,7 @@ whereBinOp op field a = WhereBinOp op nameForm sqlValue
     sqlValue = (fieldToSqlValue field a)
 
 whereValues :: [WhereForm] -> [SqlValue]
-whereValues = List.concatMap whereValues'
+whereValues = List.concatMap whereValuesInternal
 
-whereValues' :: WhereForm -> [SqlValue]
-whereValues' (WhereBinOp _ _ value) = [value]
+whereValuesInternal :: WhereForm -> [SqlValue]
+whereValuesInternal (WhereBinOp _ _ value) = [value]
--- a/src/Database/Orville/Internal/Expr/WhereExpr.hs
+++ b/src/Database/Orville/Internal/Expr/WhereExpr.hs
@@ -1,0 +1,71 @@
+module Database.Orville.Internal.Expr.WhereExpr
+( WhereExpr
+, WhereForm
+, (.==)
+, (.<>)
+, (.>)
+, (.>=)
+, (.<)
+, (.<=)
+, (%==)
+, whereValues
+) where
+
+import qualified Data.List as List
+import Data.Monoid
+import Database.HDBC
+
+import Database.Orville.Internal.Expr.Expr
+import Database.Orville.Internal.Expr.NameExpr
+import Database.Orville.Internal.FieldDefinition
+import Database.Orville.Internal.QueryKey
+import Database.Orville.Internal.Types
+
+type WhereExpr = Expr WhereForm
+
+data WhereForm
+  = WhereBinOp String NameForm SqlValue
+
+instance QualifySql WhereForm where
+  qualified (WhereBinOp op name value) table =
+    WhereBinOp op (name `qualified` table) value
+
+instance QueryKeyable WhereForm where
+  queryKey (WhereBinOp op field value) = qkOp2 op field value
+
+instance GenerateSql WhereForm where
+  generateSql (WhereBinOp op name _) =
+    (generateSql name) <> rawSql (" " <> op <> " ?")
+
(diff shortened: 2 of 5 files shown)


src/Database/Orville/Internal/Expr/NameExpr.hs
```diff
@@ -31,6 +32,9 @@ instance IsString NameForm where
 instance QualifySql NameForm where
   qualified form table = form {nameFormTable = Just table}
 
+instance QueryKeyable NameForm where
+  queryKey = QKField . nameFormName
+
```
qxjit: 🤔 Right now all the caching is done is the `QueryCache` module which only allows for selecting from one table, so it should be safe to only include the bare name here for the moment. I'm slightly concerned that we might forget about qualifying the names here if we ever allow a generic `Select` to be run with the query cache.

src/Database/Orville/Internal/Expr/WhereExpr.hs
```diff
@@ -0,0 +1,71 @@
...
+import Database.Orville.Internal.Expr.Expr
+import Database.Orville.Internal.Expr.NameExpr
+import Database.Orville.Internal.FieldDefinition
+import Database.Orville.Internal.QueryKey
+import Database.Orville.Internal.Types
+
+type WhereExpr = Expr WhereForm
+
+data WhereForm
+  = WhereBinOp String NameForm SqlValue
```
qxjit: `String` is fine for the op for this PR, but we might as well replace it with an `OpForm` that defines the operators as part of building `WhereExpr`.

src/Database/Orville/Internal/Expr/WhereExpr.hs
```diff
@@ -0,0 +1,71 @@
...
+field .< a = whereBinOp "<" field a
+
+(.<=) :: FieldDefinition a -> a -> WhereForm
+field .<= a = whereBinOp "<=" field a
+
+(%==) :: FieldDefinition a -> a -> WhereForm
+field %== a = whereBinOp "@@" field a
+
+whereBinOp :: String -> FieldDefinition a -> a -> WhereForm
+whereBinOp op field a = WhereBinOp op nameForm sqlValue
```
qxjit: I would expect this to table a `NameForm` rather than `FieldDefinition`. The whole `Expr` branch of modules should operate at a more fundamental level than the more abstract `FieldDefinition` and `TableDefinition` types. This is reflected in the module structure. Modules inside `Expr` should not "reach back out" of the `Expr` tree to grab things. Inside other items should be built on top of the `Expr` items.

src/Database/Orville/Internal/Expr/WhereExpr.hs
```diff
@@ -0,0 +1,71 @@
...
+  where
+    nameForm = fieldToNameForm field
+    sqlValue = (fieldToSqlValue field a)
+
+whereValues :: [WhereForm] -> [SqlValue]
+whereValues = List.concatMap whereValues'
+
+whereValues' :: WhereForm -> [SqlValue]
+whereValues' (WhereBinOp _ _ value) = [value]
```
qxjit: I'm not a fan of using `'` to distinguish two similar functions. I think we can do better. Do we actually need the `[WhereForm]` version of this function?

src/Database/Orville/Internal/Select.hs
```diff
@@ -73,4 +73,4 @@ rowFromSql =
     }
 
 selectField :: FieldDefinition a -> SelectForm
-selectField field = selectColumn (NameForm Nothing (fieldName field))
+selectField field = selectColumn (fieldToNameForm field)
```
qxjit: ❤️

src/Database/Orville/Internal/Where.hs
```diff
@@ -53,9 +55,7 @@ import Database.Orville.Internal.Types
   without requiring OverloadedStrings to be turned on.
 -}
 data WhereCondition
-  = forall a. BinOp String
-                    (FieldDefinition a)
-                    SqlValue
+  = WhereConditionForm E.WhereForm
```
qxjit: 👍

src/Database/Orville/Internal/Where.hs
```diff
@@ -86,37 +86,39 @@ instance QueryKeyable WhereCondition where
...
 
 whereConditionSql :: WhereCondition -> String
 whereConditionSql cond = internalWhereConditionSql Nothing cond
 
 internalWhereConditionSql ::
      Maybe (TableDefinition a b c) -> WhereCondition -> String
-internalWhereConditionSql tableDef (BinOp op fieldDef _) =
-  qualifiedFieldName tableDef fieldDef ++ " " ++ op ++ " ?"
+internalWhereConditionSql (Just tableDef) (WhereConditionForm form) =
+  rawExprToSql . generateSql $ form `qualified` (tableName tableDef)
```
qxjit: ❤️  It feels so good use `qualified` here instead of the one that is specific to this module right now.

src/Database/Orville/Internal/Expr/NameExpr.hs
exodrifter: We should do this now rather than later. Would it be fine to use `unescapedName`, or would `qkOp2` be better?

src/Database/Orville/Internal/Expr/WhereExpr.hs
exodrifter: I've created ch12473 to track this issue

src/Database/Orville/Internal/Expr/WhereExpr.hs
exodrifter: We will need the `[WhereForm]` version of this function in several places:

* Database.Orville.Core
`deleteWhereBuild` [link](https://github.com/flipstone/orville/blob/4f2dd5aa31f625829a668e349ee80ea2d4370c4b/src/Database/Orville/Core.hs#L200)
`updateFields` [link](https://github.com/flipstone/orville/blob/4f2dd5aa31f625829a668e349ee80ea2d4370c4b/src/Database/Orville/Core.hs#L251)
* Database.Orville.Internal.SelectOptions
`selectOptValues` [link](https://github.com/flipstone/orville/blob/4f2dd5aa31f625829a668e349ee80ea2d4370c4b/src/Database/Orville/Internal/SelectOptions.hs#L96)

I'll change the name to `whereValuesInternal`

src/Database/Orville/Internal/Expr/WhereExpr.hs
qxjit: I would consider `SqlValue` ever more low lever than `Expr`. Although it might be nice to have `Expr` be free of dependencies on the HDBC library, doing so would require an alternate representation of values that are going to be filled in for placeholders, which would likely end up looking a lot like `SqlValue`.

For the record, I wouldn't mind seeing us replace `SqlValue` at some point globally anyhow, but that's a much bigger undertaking and should be probably be handled with the driver extraction and moving over to native postgres library work that will come later.

PR #25, opened 2018-07-19
Add whereRaw to Where module

exodrifter:
Adds the `whereRaw` function to the the `Database.Orville.Internal.Where` module. This will allow a user of the `Where` API to create where conditions using raw strings.

In order to make this work, `WhereConditionForm` has been renamed to `WhereConditionExpr` and now contains a `WhereExpr` instead of a `WhereForm`.

### DIFF
--- a/test/WhereConditionTest.hs
+++ b/test/WhereConditionTest.hs
@@ -9,6 +9,7 @@ import qualified TestDB as TestDB
 
 import Control.Monad (void)
 import Data.Int (Int64)
+import Database.HDBC (toSql)
 import Database.Orville ((.==))
 import Database.Orville.Expr (aliased, qualified)
 import Test.Tasty (TestTree, testGroup)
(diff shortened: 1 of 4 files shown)


test/WhereConditionTest.hs
```diff
@@ -68,6 +68,20 @@ test_where_condition =
...
             result
+      , testCase "Raw where" $ do
+          run (TestDB.reset schema)
+          void $ run (O.insertRecord orderTable foobarOrder)
+          void $ run (O.insertRecord orderTable orderNamedAlice)
+          void $ run (O.insertRecord customerTable aliceCustomer)
+          void $ run (O.insertRecord customerTable bobCustomer)
+          let opts =
+                O.where_ $
+                O.whereRaw $ "customer.name = 'Alice'"
```
qxjit: oo... this brings up a really good question. `whereRaw` probably needs to support placeholders to be used with `SqlValue` so that users don't have to engage in dangerous string concatenation to put comparison values into their where conditions!

exodrifter: oh dear

PR #26, opened 2018-07-20
Move In and NotIn to WhereExpr

exodrifter:
Move the `whereIn` and `whereNotIn` functions to the `WhereExpr` module.

This also fixes a few potential bugs:
* Fixes not using `List.nub` on the `[SqlValue]` argument for `whereNotIn` 
* Fixes not using `AlwaysTrue` when empty array is passed to `whereNotIn`
* Fixes `whereIn` and `(.<-)` having different behaviour

### DIFF
--- a/src/Database/Orville/Internal/Expr/WhereExpr.hs
+++ b/src/Database/Orville/Internal/Expr/WhereExpr.hs
@@ -8,7 +8,10 @@ module Database.Orville.Internal.Expr.WhereExpr
 , (.<)
 , (.<=)
 , (%==)
+, (.<-)
 , whereValues
+, whereIn
+, whereNotIn
 , whereNull
 , whereNotNull
 , whereRaw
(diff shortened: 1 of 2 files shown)


src/Database/Orville/Internal/Expr/WhereExpr.hs
```diff
@@ -75,6 +102,17 @@ name .<= value = WhereBinOp "<=" name value
...
 
+(.<-) :: NameForm -> [SqlValue] -> WhereForm
+name .<- values = whereIn name values
+
+whereIn :: NameForm -> [SqlValue] -> WhereForm
+whereIn _ [] = WhereAlwaysFalse
+whereIn field values = WhereIn field (List.nub values)
+
+whereNotIn :: NameForm -> [SqlValue] -> WhereForm
+whereNotIn _ [] = WhereAlwaysTrue
```
qxjit: Separate from this PR, I do think it's a good idea to bump up our `base` constraint so that we can use non-empty lists here and avoid these things. I have a feeling we might hit a couple of bugaboos in the Popper code where we have to handle empty lists in that case, but it would be worth it in the long run.

exodrifter: Made ch12475 to track this issue

PR #28, opened 2018-08-23
Use current schema instead of public

exodrifter:
By default, postgres creates new indexes under the schema given by `current_schema()`, but Orville checks for the existence of an index in the `public` schema. This PR fixes this discrepancy by making the check use `current_schema()`.

### DIFF
--- a/src/Database/Orville/Internal/MigrateConstraint.hs
+++ b/src/Database/Orville/Internal/MigrateConstraint.hs
@@ -47,6 +47,6 @@ getConstraints conn = do
       "SELECT conname \
                         \FROM pg_constraint \
                         \JOIN pg_namespace ON pg_namespace.oid = pg_constraint.connamespace \
-                        \WHERE nspname = 'public'"
+                        \WHERE nspname = current_schema()"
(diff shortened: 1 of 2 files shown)


src/Database/Orville/Internal/MigrateConstraint.hs
```diff
@@ -47,6 +47,6 @@ getConstraints conn = do
       "SELECT conname \
                         \FROM pg_constraint \
                         \JOIN pg_namespace ON pg_namespace.oid = pg_constraint.connamespace \
-                        \WHERE nspname = 'public'"
+                        \WHERE nspname = 'current_schema()'"
```
qxjit: I'm guessing you didn't want this to be in the single quotes here?

If no tests failed, then we probably need to make sure we have a constraint in our schemas to ensure this is caught.

exodrifter: Oh no.

The tests didn't fail, but I don't think we have any tests for indices yet, much less tests for different schemas.

PR #30, opened 2018-08-29
Show the value that could not be decoded

exodrifter:
When throwing a `RowDataError`, Orville does not include what the value that could not be decoded was.

This makes the error useless when trying to understand why something broke in a production environment, especially when a database query doesn't reveal any immediately obvious solutions. Including the value that could not be decoded can make similar situations easier to debug.


exodrifter:
This has the possibility of exposing potentially sensitive data. There might be a better way to make the error message more useful. For example, only including the row id of the column we failed to decode.

PR #33, opened 2018-09-18
[ch12596] Moves the startTransactionSQL callback into OrvilleEnv

qxjit:
This simplifies the typeclass and opens the door to more sophisticated
callbacks being registered in the OrvilleEnv datatype rather than flooding
the typeclass with many methods.

It's possible this is the beginning of a "Driver"-like framework, but it's
unclear. Mostly I want to see if this is workable from an end-user standpoint.

### DIFF
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -37,6 +37,7 @@ module Database.Orville.Core
   , readOnlyField
   , OrvilleEnv
   , newOrvilleEnv
+  , setStartTransactionSQL
   , ormEnvPool
   , Orville
   , OrvilleT
--- a/src/Database/Orville/Core.hs
+++ b/src/Database/Orville/Core.hs
@@ -39,7 +39,9 @@ module Database.Orville.Core
   , newOrvilleEnv
   , setStartTransactionSQL
   , aroundRunningQuery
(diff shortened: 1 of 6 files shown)


qxjit:
@jbrechtel This is really about the use case of `startTransactionSQL` in itself, but rather altering the pattern of how one engages with "lower level" hooks in orville. My hope is that rather than having to add every single hook into the typeclass, registering them in the `OrvilleEnv` provides a more palatable and gradual-introduction friendly interface.

I'm shortly going to try adding hooks to the transaction code that will enable the trigger module to decide when to emit triggers based on what is happening in the transaction code (i.e. either immediately, on commit, or discarded on rollback). Adding those hooks into the typeclass feels somehow gross to me.

As a pratical pattern, implementing them at the typeclass level *ergonomically* requires us to provide a default, but the default implementation (no-op) would *fail* to properly wrap the trigger behavior, forcing the user to be aware of these callbacks in order to use the trigger module. I'm also hoping that pushing them into the env will enable me to both keep it hidden *and* make the default behavior the correct behavior.

cmika:
We were thinking about adding a monadic check but haven't implemented anything like that yet. We are currently always using the same transaction text.

qxjit:
@cmika 👍

PR #34, opened 2018-09-19
[ch12596] Moves runningQuery callback into OrvilleEnv

qxjit:
test/TestDB.hs provides an instructive example of the impact of
this change to users of the tracing feature. On the one hand I
think the user code is more straightforward overall, with the
tracing concern now being concentrated in just a couple of
functions. On the other hand, the tracing concern has been
completely eliminated from the type signatures, so the type
system cannot be used to make guarantees about it.

### DIFF
--- a/src/Database/Orville/Internal/Monad.hs
+++ b/src/Database/Orville/Internal/Monad.hs
@@ -45,11 +45,18 @@ data QueryType
  one.
 -}
 data OrvilleEnv conn = OrvilleEnv
-  { ormEnvPool :: Pool conn
-  , ormEnvConnectionEnv :: Maybe (ConnectionEnv conn)
+  { ormEnvConnectionEnv :: Maybe (ConnectionEnv conn)
   , ormEnvStartTransactionSQL :: String
   , ormEnvRunningQuery :: forall a. QueryType -> String -> IO a -> IO a
+  , ormEnvTransactionCallback :: TransactionEvent -> IO ()
+  , ormEnvPool :: Pool conn
   }
+
+data TransactionEvent
+  = TransactionStart
+  | TransactionCommit
+  | TransactionRollback
+  deriving (Ord, Eq, Enum, Show, Read)
 
 defaultStartTransactionSQL :: String
 defaultStartTransactionSQL = "START TRANSACTION"
@@ -59,6 +66,9 @@ setStartTransactionSQL sql env = env {ormEnvStartTransactionSQL = sql}
 
 defaultRunningQuery :: QueryType -> String -> IO a -> IO a
 defaultRunningQuery _ _ action = action
+
+defaultTransactionCallback :: TransactionEvent -> IO ()
+defaultTransactionCallback = const (pure ())
 
 aroundRunningQuery ::
      (forall a. QueryType -> String -> IO a -> IO a)
@@ -71,6 +81,15 @@ aroundRunningQuery outside env = env {ormEnvRunningQuery = layeredAround}
       outside queryType sql (inside queryType sql action)
     inside = ormEnvRunningQuery env
 
+addTransactionCallBack ::
+     (TransactionEvent -> IO ()) -> OrvilleEnv conn -> OrvilleEnv conn
+addTransactionCallBack callback env =
+  env {ormEnvTransactionCallback = wrappedCallback}
+  where
+    wrappedCallback event = do
+      ormEnvTransactionCallback env event
+      callback event
+
 {-|
  'newOrvilleEnv' initialized an 'OrvilleEnv' for service. The connection
  pool provided will be used to obtain connections to the database ase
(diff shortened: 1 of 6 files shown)


src/Database/Orville/Internal/Monad.hs
```diff
@@ -56,14 +57,29 @@ defaultStartTransactionSQL = "START TRANSACTION"
...
+defaultRunningQuery :: QueryType -> String -> IO a -> IO a
+defaultRunningQuery _ _ action = action
+
+aroundRunningQuery ::
+     (forall a. QueryType -> String -> IO a -> IO a)
+  -> OrvilleEnv conn
+  -> OrvilleEnv conn
+aroundRunningQuery outside env = env {ormEnvRunningQuery = layeredAround}
+  where
+    layeredAround, inside :: QueryType -> String -> IO a -> IO a
```
jbrechtel: that's new syntax for me

qxjit: It doesn't come up very often, but it's occasionally nice to only write the type signature once when multiple values have the same signature.

qxjit:
@jbrechtel Yes, with that naming I would trying to capture the fact that it's layered on top of whatever else is already in the env. I'd certainly entertain a new name to reflect that. I wasn't totally happy with it as is.

qxjit:
@jbrechtel @telser The way `aroundRunningQuery` is written makes it *impossible* to change the query tracing behavior wholesale. It *only* allows for layering on. Do you think this is a good thing, or a bad thing? My current assumption it's a good thing because there might be a callback in there that you don't know about and might accidentally through away...

Specifically, I find myself facing the same decision point for how to add a new callback to listen for transaction start, commit, rollbacks. I can either wrap it around whatever existing callback is in the `env`, *or* throw the existing one away.  I'm inclined to do the former so that someone using the new trigger module (which will use that callback) doesn't accidentally through away the trigger callbacks by adding their own callback.

src/Database/Orville/Internal/Monad.hs
```diff
@@ -56,14 +57,29 @@ defaultStartTransactionSQL = "START TRANSACTION"
...
+    layeredAround queryType sql action =
+      outside queryType sql (inside queryType sql action)
```
telser: Since there is already discussion around names.. Is sql here really a raw sql string? Just going by the types. And if so, should it be, or would it be at all helpful to force a more expressive type for this?

test/TestDB.hs
```diff
@@ -25,39 +23,44 @@ import qualified Database.Orville.Raw as ORaw
 
 type TestPool = Pool Postgres.Connection
 
-type QueryWriterT = WriterT [(O.QueryType, String)]
+data Trace = Trace
```
telser: This won't cause a conflict with something from Debug.Trace, correct? I don't _think_ it does, but haven't actually had to use it in forever.

test/TestDB.hs
```diff
@@ -25,39 +23,44 @@ import qualified Database.Orville.Raw as ORaw
...
-  { runTestMonad :: QueryWriterT (O.OrvilleT Postgres.Connection IO) a
+  { runTestMonad :: O.OrvilleT Postgres.Connection IO a
   } deriving ( Functor
              , Applicative
              , Monad
              , MonadIO
              , MonadBase IO
              , MonadThrow
              , MonadCatch
+             , O.MonadOrville Postgres.Connection
```
telser: I find it interesting this becomes derivable now.

test/TestDB.hs
```diff
@@ -25,39 +23,44 @@ import qualified Database.Orville.Raw as ORaw
...
+queryTrace queryPred action = do
+  ref <- liftIO (newIORef [])
+  void $ O.localOrvilleEnv (addTraceToEnv (Trace queryPred ref)) action
+  reverse <$> liftIO (readIORef ref)
+
+addTraceToEnv :: Trace -> O.OrvilleEnv conn -> O.OrvilleEnv conn
+addTraceToEnv trace =
+  O.aroundRunningQuery $ \queryType sql action -> do
+    when
+      (tracePred trace queryType)
```
telser: This is a style question: Upon hitting this code I immediately misread the `when` as `where` most likely from the newline. Does anyone else ever have that issue? It's the reason for my preference of
``when bool\n a`` (I'm terrible at markdown, but hopefully it gets the point across).

test/TestDB.hs
```diff
@@ -25,39 +23,44 @@ import qualified Database.Orville.Raw as ORaw
...
+      (modifyIORef (traceRef trace) ((queryType, sql) :))
```
telser: Do we want `modifyIORef'` here similar to usually preferring the prime versions of `foldl`?

src/Database/Orville/Internal/Monad.hs
qxjit: It would have to be a newtype around `String` (rather than a structured type)  because this is at the lowest level of "hey, I'm actually running this sql", so it could be *any* sql whatsover (even raw sql provided by the user). I'm open newtyping it, but that certainly beyond the scope of my goals here.

test/TestDB.hs
qxjit: I'm not too concerned about it since it's in the test suite.

test/TestDB.hs
qxjit: I would gladly format it that way, but right now all the formatting is done by `hindent`....

test/TestDB.hs
qxjit: Probably just by way of setting a good example, yeah. I wasn't thinking at all about performance when doing this since it's just in the tests.

test/TestDB.hs
qxjit: Yeah, that felt good for sure. If only we could get rid of `MonadBaseControl` now....

jbrechtel:
@qxjit  re "It only allows for layering on." I feel OK with this but it may be a lack of imagination. The main thing that comes to mind is something that says "I just want to log everything but not execute it." or something along those lines. But there are much better ways to achieve that for most cases I'd want that behavior anyway.

test/TestDB.hs
qxjit: @telser I'll address this once I've merged everything, just to keep a clean history.

PR #35, opened 2018-09-19
 [ch12596] Adds TransactionEvent broadcasting

qxjit:
This adds a callback to allow Orville to let interested parties know
when transactions start,commit, and rollback. The callbacks are
monomorphized to IO for simplicity and clarity. The presumption is
that if you are a sophisticated enough user to be interested in these
callbacks that you will understand the value of dealing with any
necessary unlifting yourself rather than having Orville tie you to a
specific unlifting implementation.

This does, however, eliminate the possibility of using a stateful
monad (Writer, State) if your goal is simply to record a list of the
transaction events that happen. Instead you have to use an IORef (or
similar) construct that you can access from the IO monad.

### DIFF
--- a/src/Database/Orville/Internal/Monad.hs
+++ b/src/Database/Orville/Internal/Monad.hs
@@ -45,11 +45,18 @@ data QueryType
  one.
 -}
 data OrvilleEnv conn = OrvilleEnv
-  { ormEnvPool :: Pool conn
-  , ormEnvConnectionEnv :: Maybe (ConnectionEnv conn)
+  { ormEnvConnectionEnv :: Maybe (ConnectionEnv conn)
   , ormEnvStartTransactionSQL :: String
   , ormEnvRunningQuery :: forall a. QueryType -> String -> IO a -> IO a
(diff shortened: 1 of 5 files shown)


src/Database/Orville/Internal/Monad.hs
```diff
@@ -71,15 +81,28 @@ aroundRunningQuery outside env = env {ormEnvRunningQuery = layeredAround}
...
     inside = ormEnvRunningQuery env
 
+addTransactionCallBack ::
+     (TransactionEvent -> IO ()) -> OrvilleEnv conn -> OrvilleEnv conn
+addTransactionCallBack callback env =
+  env {ormEnvTransactionCallback = wrappedCallback}
+  where
+    wrappedCallback event = do
+      ormEnvTransactionCallback env event
+      callback event
```
qxjit: This wraps any existing callback similar to `aroundRunningQuery`. This does of course mean that exceptions that occur in the first callback nullify the second one. I hate callbacks 😢

PR #36, opened 2018-09-21
[ch12596] [WIP] A first cut at triggers respecting transactions. Not to be merged

qxjit:
This successfully implements a crude TriggerT monad transformer that supports using
triggered versions of the insert/update/delete functions. Aside from the flippant
use of IORefs (rather than carefully considered selection of a concurrency primitive),
there are a couple of other issues with this implementation:

* Any monadic state modifications made by the trigger callback *during a transaction commit*
  are thrown away, due to the fact that the transaction callbacks are monomorphised to IO.
  This will need to be re-visited

* It is possible to use `runTriggerT` to layer on triggering *in the middle* of another
  *untriggered* Orville computation. This is nice flexbility, but if the outer computation
  has a transaction open, the inner triggered computation does not currently know this,
  nor would it ever see the commit of the transaction inside the scope of its own computation.
  It may be that this is fundamentally unsound, or possibly requires extreme weirdness to
  make work (e.g. saving the triggers and running them later outside the triggered computation
  context when the commit happens????)

If you have time, please take a look and provide any thoughts or feedback you might have.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -58,6 +58,7 @@ library:
     - Database.Orville.PostgresSQL
     - Database.Orville.Raw
     - Database.Orville.Select
+    - Database.Orville.Trigger
 tests:
   spec:
     main: Driver.hs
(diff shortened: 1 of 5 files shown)


qxjit:
Closing this after discussion with @telser. I'm going to reduce the scope of what `TriggerT` allows you to do and try again in another PR.

PR #37, opened 2018-09-26
[ch12596] [WIP] A second cut at triggers respecting transactions. Not to be merged

qxjit:
This reduces the scope of the TriggerT transformer (now OrvilleTriggerT) from the previous
PR in two important ways:

1) It includes the OrvilleT directly, rather than allowing TriggerT to be layered on top of
   any old MonadOrville implementation. This allows runOrvilleTriggerT to have control over
   when the OrvilleEnv is instantiated to avoid runOrvilleTriggerT being called with a transaction
   already in progress.

2) It only allows for keeping track of a list of triggers as data to be processed at a later
   time, rather than executing monadic trigger actions in line. This removed the need for any
   runInBase nonsense because all the action performed by runOrvilleTriggerT can be easily
   monomorphised to IO

This in no way reduced the scope of the `MonadTrigger` class itself, only the scope of the
default implementation. If a specific application wanted different behavior, it could implement
a custom `MonadTrigger` implementation and still use the provided insert/update/delete functions
for sending the triggers.

This PR does *not* address using an appropriate data structure for efficient appending and
iteration of triggers. I will addresss such issues once I am happy with the Monadic structures
involved in the computation.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -58,6 +58,7 @@ library:
     - Database.Orville.PostgresSQL
     - Database.Orville.Raw
     - Database.Orville.Select
+    - Database.Orville.Trigger
 tests:
   spec:
     main: Driver.hs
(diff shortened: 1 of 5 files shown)


liminalisht:
This looks really great to me.

qxjit:
Closing this. Will produce a non-WIP PR and post the link to it here.

qxjit:
This was reposted as https://github.com/flipstone/orville/pull/38

PR #38, opened 2018-10-01
[ch12596] Adds Database.Orville.Trigger

qxjit:
This provides OrvilleTriggerT which can be used instead of OrvilleT.
With OrvilleTriggerT you can use the insertTriggered, updateTriggered
and deleteTriggered operations to have OrvilleTriggerT automatically
keep track of an application specific list of trigger events that
occurred in while running a block. The list of trigger events can then
be used for later processing, such as scheduling background jobs to
do extra processing in response to them.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -58,6 +58,7 @@ library:
     - Database.Orville.PostgresSQL
     - Database.Orville.Raw
     - Database.Orville.Select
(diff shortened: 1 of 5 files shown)


qxjit:
This is a repost of https://github.com/flipstone/orville/pull/37

PR #40, opened 2018-10-12
Give RelationalMap a Profunctor instance.

liminalisht:
This PR gives `RelationalMap` a `Profunctor` instance. `dimap f g` is arguably clearer than `O.mapAttr f . fmap g` for those who understand what a `Profunctor` is.

### DIFF
--- a/src/Database/Orville/Internal/RelationalMap.hs
+++ b/src/Database/Orville/Internal/RelationalMap.hs
@@ -25,7 +25,7 @@ module Database.Orville.Internal.RelationalMap
 import Control.Monad (join, when)
 import Control.Monad.Reader (ask)
 import Control.Monad.State (modify)
-import Data.Profunctor(Profunctor(..))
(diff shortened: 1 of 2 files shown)


src/Database/Orville/Internal/RelationalMap.hs
```diff
@@ -25,6 +25,7 @@ module Database.Orville.Internal.RelationalMap
 import Control.Monad (join, when)
 import Control.Monad.Reader (ask)
 import Control.Monad.State (modify)
+import Data.Profunctor(Profunctor(..))
```
qxjit: I would even explicitly import `rmap` and `lmap` here rather than use open imports. It will never make a practical difference really, but it's still nice to see the function names in the import list for maximum clarity. Especially since we're not actually using *all* the methods of `Profunctor` here.

liminalisht: Yes, I'll do this

PR #46, opened 2018-11-06
Adds travis-ci builds for lts-12 and ghc8.4.4

telser:
Not very familiar with how this works, but hopefully this does it and gets us some testing with more recent GHC/lts versions.

I got the cabal version from the upload time at https://launchpad.net/~hvr/+archive/ubuntu/ghc/+index?batch=75&memo=75&start=75 and checked that it is the version installed on a lts-12.x project.


.travis.yml
```diff
@@ -95,6 +99,11 @@ matrix:
   # - env: BUILD=stack ARGS="--resolver lts-8"
   #   compiler: ": #stack 8.0.2 (lts-8)"
   #   addons: {apt: {packages: [libgmp-dev]}}
+  
+  # conduit dependency keeps this from working right now, so disabling
+  #- env: BUILD=stack ARGS="--resolver lts-12"
+  #  compiler: ": #stack 8.4.4 (lts-12)"
+  #  addons: {apt: {packages: [libgmp-dev]}}
```
qxjit: @telser Can you elaborate on this? Is it because conduit is not in `lts-12` at all, or because we have a conduit version constraint problem?

qxjit: @telser @liminalisht I see there is already a branch to bump to conduit version. Should this be merged and would it fix the `lts-12` problem?  https://github.com/flipstone/orville/commit/c8de554b71c7bf267076d1f0e1bd89a6fa46b3fc

telser: We don't want to merge it because there are API changes in conduit 1.3 that we need to address first. I've started looking at it some.

PR #47, opened 2018-11-07
liminalisht/timefix

liminalisht:
This PR adds a crucial fix to the way sql values of type `timestamp with time zone` are parsed into `UTCTime`.

You are likely to get this sort of error when trying to read a row with a column of type `timestamp with time zone`: 

```
RowDataError \"Error decoding data from column active_at value\""
```

This code change seems to fix the issue.

### DIFF
--- a/src/Database/Orville/Internal/SqlType.hs
+++ b/src/Database/Orville/Internal/SqlType.hs
@@ -358,7 +358,8 @@ utcTimeToSql = HDBC.SqlUTCTime
 utcTimeFromSql :: HDBC.SqlValue -> Maybe Time.UTCTime
 utcTimeFromSql sql =
   case sql of
-    HDBC.SqlUTCTime t -> Just t
+    HDBC.SqlUTCTime   t -> Just t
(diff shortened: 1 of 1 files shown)


src/Database/Orville/Internal/SqlType.hs
```diff
@@ -358,7 +358,8 @@ utcTimeToSql = HDBC.SqlUTCTime
 utcTimeFromSql :: HDBC.SqlValue -> Maybe Time.UTCTime
 utcTimeFromSql sql =
   case sql of
-    HDBC.SqlUTCTime t -> Just t
+    HDBC.SqlUTCTime   t -> Just t
+    HDBC.SqlZonedTime t -> Just (Time.zonedTimeToUTC t)
```
qxjit: @liminalisht I'm fine with you merging this. We should also add a test to so we don't break it again in the future.

liminalisht: :+1:

PR #50, opened 2018-11-13
Unshackles MonadOrville from MonadBaseControl

qxjit:
This replaces the MonadBaseControl superclass constraint for MonadOrville
with a constraint for a new MonadOrvilleControl typeclass instead.
MonadOrvilleControl defines the specific control lifting functions that
Orville needs rather than requiring the full power of either MonadBaseControl
or MonadUnliftIO.

An unexpected outcome of this was that much of the testing infrastructure was
able to abandon MonadBaseControl immediately, without replacing it with
MonadUnliftIO at all. This is mainly due to the fact that Orville provides a
ReaderT instance for MonadOrvilleControl that doesn't rely on any extrenal
libraries. This only works because the tests only use ReaderT in their Monad
stacks. In the real world, users will have other transformers (e.g. for
web frameworks) that will need help providing instances of MonadOrvilleControl.
We can build a library of helper functions in Orville for implementing
MonadOrvilleControl based on commonly used libraries, both general purpose
and even web-framework specific (in their own libraries).

The existing MonadBaseControl dependencies in the library entirely moved to
Database.Orville.MonadBaseControl which is not exported from Core. Instaed,
users can import this module explicitly if they know that they need to use
MonadBaseControl.

Also note: `hindent` has mangled some comment formatting a bit. I'm planning
to move away from `hindent` (as as result of this and other frustrations). I could
not find a good solution to fix any of the comment format weirdness.

### DIFF
--- a/src/Database/Orville/Internal/Monad.hs
+++ b/src/Database/Orville/Internal/Monad.hs
@@ -3,11 +3,9 @@ Module    : Database.Orville.Internal.Monad
 Copyright : Flipstone Technology Partners 2016-2018
 License   : MIT
 -}
-{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Database.Orville.Internal.Monad where
@@ -17,8 +15,6 @@ import Control.Monad.Base
 import Control.Monad.Catch (MonadCatch, MonadMask(..), MonadThrow)
 import Control.Monad.Except
 import Control.Monad.Reader
-import Control.Monad.State
-import Control.Monad.Trans.Control
 import Data.Pool
 import Database.HDBC hiding (withTransaction)
 
@@ -132,7 +128,7 @@ withConnectionEnv action = do
   case ormEnvConnectionEnv ormEnv of
     Just connected -> action connected
     Nothing ->
-      withResource (ormEnvPool ormEnv) $ \conn -> do
+      liftWithConnection (withResource (ormEnvPool ormEnv)) $ \conn -> do
         let connected = newConnectionEnv conn
         localOrvilleEnv (const $ ormEnv {ormEnvConnectionEnv = Just connected}) $
           action connected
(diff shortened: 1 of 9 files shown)


src/Database/Orville/MonadBaseControl.hs
```diff
@@ -0,0 +1,116 @@
...
+  |-}
+instance MTC.MonadTransControl (O.OrvilleT conn) where
+  type StT (O.OrvilleT conn) a = MTC.StT (ReaderT (O.OrvilleEnv conn)) a
+  liftWith = MTC.defaultLiftWith Internal.OrvilleT Internal.unOrvilleT
+  restoreT = MTC.defaultRestoreT Internal.OrvilleT
+
+{-|
+   Because we recommend using 'MonadUnliftIO' rather than 'MonadBaseControl',
+   we do not provide 'MonadBaseControl' instance for 'OrvilleT' by default
+   along wit the definition. If you do need to use 'MonadBaseControl',
```
telser: wit -> with

src/Database/Orville/Internal/Monad.hs
```diff
@@ -181,17 +250,22 @@ instance MonadOrville conn m => MonadOrville conn (ReaderT a m) where
...
-instance MonadOrville conn m => MonadOrville conn (StateT a m) where
-  getOrvilleEnv = lift getOrvilleEnv
-  localOrvilleEnv modEnv action =
-    StateT $ \val -> localOrvilleEnv modEnv (runStateT action val)
-
-instance MonadTransControl (OrvilleT conn) where
-  type StT (OrvilleT conn) a = StT (ReaderT (OrvilleEnv conn)) a
-  liftWith = defaultLiftWith OrvilleT unOrvilleT
-  restoreT = defaultRestoreT OrvilleT
+-- ReaderT is trivial enough that we just provide a 'MonadOrvilleControl'
```
telser: I like this approach. Not requiring, but allowing use of either `MonadUnliftIO` or `MonadBaseControl` feels like a big win.

src/Database/Orville/Internal/Monad.hs
```diff
@@ -3,11 +3,9 @@ Module    : Database.Orville.Internal.Monad
 Copyright : Flipstone Technology Partners 2016-2018
 License   : MIT
 -}
-{-# LANGUAGE FlexibleContexts #-}
```
telser: It's awesome that the more "core" module got to drop some language extensions in this work.

src/Database/Orville/MonadBaseControl.hs
qxjit: good catch. Thanks

src/Database/Orville/Internal/Monad.hs
qxjit: Agreed. Initially I thought that the added complexity for end users would be too high, but in the end I'm quite happy with the tradeoff.

src/Database/Orville/Internal/Monad.hs
qxjit: Agreed.

qxjit:
@telser To your questions:

1) We definitely will provide helpers for `MonadUnliftIO`, it's just beyond the scope of this PR. 

2) I love the idea of breaking things ups in general, but I'm pessimistic about it in this case because
    supporting MBC requires access to the internal constructors for `OrvilleT`. While we could split
   out the helper functions, we would not be able to provide a `MonadBaseControl` instance for     
   `OrvilleT` if it were in a separate package.

3) Yes. I think we should have a least something that verifies both `MonadBaseControl` and `MonadUnliftIO` work. I plan to do that as part of the more general `MonadUnliftIO` work.

PR #51, opened 2018-11-16
Adds support for MonadUnliftIO

qxjit:
Rather than making this explicitly the default in the library by
exporting the UnliftIO oriented functions from the Core module, I
opted instead to built it as a peer to MonadBaseControl. This does
not represent a change in thinking that we should direct new users to
use UnliftIO as a default approach, merely that we should convey that
opinion through means other that module organization in the library.
For instance, doing it through both comments in the API documentation
and directing users to UnliftIO rather than MonadBaseControl in any
tutorial docs we write in the future.

### DIFF
--- a/orville.cabal
+++ b/orville.cabal
@@ -2,7 +2,7 @@
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: c5fd7779b4d320abcf36b279057892f7585c711e87e5008c3edf3c09d21eafd7
+-- hash: 62042f25d9401cb71f84fe356b1e1a883494786af00ac6c0faae07323f16a9e4
 
 name:           orville
 version:        0.8.3.0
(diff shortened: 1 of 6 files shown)


telser:
So something I don't like about this is it means all users of Orville must pull in MonadUnliftIO and MBC even if they only use one.

qxjit:
@telser Philosophically I agree. Practically speaking there are a number of factors I see:

* We don't currently know how aggressively we want to break Orville into multiple packages, including whatever potential confusion that causes for new users
* We don't currently know the mechanics of how we would want such a breakup to work
* Moving either these instances or the `MonadBaseControl` ones to another library would involve breaking some amount of encapsulation around `OrvilleT`
* The dependencies of `unliftio-core` are truly minimal (`base` and `transformers`)
* The dependencies of `monad-control` are slightly less minimal (`stm`, `transformers-base`, `transformers-compat`
* It doesn't force users to actually depend on these libraries themselves if they don't use them, so if the dependencies are minimal it may not be worth whatever cost we/they pay to break them out.

That's all to say that I think making a call about this is premature right now, and certainly outside the scope of this PR. I definitely agree it's something that deserves more though before we bless an `0.9` release.

PR #52, opened 2018-11-27
Extracts HasOrvilleContext typeclass from MonadOrville.

qxjit:
Originally I was hoping to simply invert the MonadOrvilleControl ->
MonadOrville relationship so that the the functions for managing context were
in the superclass and control functions in the subclass. That failed, however,
because then no MonadOrvilleControl instance can exist for IO, because IO lacks
any Orville context to manage.

Instead, I ended up extracting the context management functions to
HasOrvilleContext and changing MonadOrville to an empty typeclass that has
superclass contraints for all the classes that Orville functions need. This
allows the context management functions and the control lifting functions to be
implemented by different layers of the monad stack, and by different lifting
strategies. In the end, only the top layer of the monad stack is required to
actually implement *both*, by providing a MonadOrville instance.

In practice, this means that there are *three* typeclasses that a user needs to
be aware of, which is someone annoying. However, `HasOrvilleContext` should
almost always be derivable and `MonadOrville` instances are trivial, so I think
this cost is a reasonable tradeoff to provide the flexibility to use either
`MonadUnliftIO` or `MonadBaseControl` (or another solution) as required. For
release, we will just need to make sure we have enough documentation and
tutorials in place to satisfy users.

### DIFF
--- a/src/Database/Orville/Internal/Monad.hs
+++ b/src/Database/Orville/Internal/Monad.hs
@@ -14,7 +14,8 @@ import Control.Applicative
 import Control.Monad.Base
 import Control.Monad.Catch (MonadCatch, MonadMask(..), MonadThrow)
 import Control.Monad.Except
-import Control.Monad.Reader
+import Control.Monad.Reader (ReaderT(..), ask, local, mapReaderT, runReaderT)
+import Control.Monad.State (StateT, mapStateT)
 import Data.Pool
 import Database.HDBC hiding (withTransaction)
 
@@ -35,7 +36,7 @@ data QueryType
  'OrvilleEnv' tracks all the environment information required for an
  'OrvilleT conn m' Monad to operate. Use 'newOrvilleEnv' to construct
  one.
--}
+  -}
 data OrvilleEnv conn = OrvilleEnv
   { ormEnvConnectionEnv :: Maybe (ConnectionEnv conn)
   , ormEnvStartTransactionSQL :: String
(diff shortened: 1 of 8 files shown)


src/Database/Orville/Internal/Monad.hs
```diff
@@ -189,12 +186,43 @@ class ( Monad m
       liftWithConnection = defaultLiftWithConnection wrapMyMonad unWrapMyMonad
       liftFinally = defaultLiftFinally wrapMyMonad unWrapMyMonad
    @
- |-}
+
+   If you are using transformers in your monad stack other beyond 'ReaderT',
```
telser: I think there is a typo in "other beyond" :)

src/Database/Orville/Internal/Monad.hs
```diff
@@ -204,7 +232,7 @@ instance MonadOrvilleControl IO where
    'liftWithConnection' for 'MonadOrvilleControl' instances when the Monad in
    question is a wrapper around a type that already implements
    'MonadOrvilleControl'
-  |-}
+  -}
```
telser: IIRC haddock doesn't need the `|`, so is that the reason for the removal here?

qxjit: Yeah. It looks like I got in the habit of using two `|` in these comments at some point, but haddock only cares about the first ones. I'm going to keep cleaning these up as a run across them.

src/Database/Orville/Internal/Monad.hs
qxjit: Good catch. I deleted the word "other" in my head after typing it, but never actually deleted it from the comment.

PR #59, opened 2019-03-19
[ch14845] Add Unbounded Text column type

onslaughtq:
An unbounded text column type is added to Orville.  Notably, for postgresql, this can either be a Text column type, or a Varchar Column, without a length specified.  Neither is an ANSI standard, and I'm not entirely certain how strict orville needs to be here, especially this older development version.


src/Database/Orville/Internal/MigrateTable.hs
```diff
@@ -149,6 +150,7 @@ columnTypeSqlId BigInteger = SqlBigIntT
 columnTypeSqlId Double = SqlFloatT
 columnTypeSqlId (VarText _) = SqlVarCharT
 columnTypeSqlId (Text _) = SqlCharT
+columnTypeSqlId UnboundedText = SqlVarCharT
```
qxjit: I think this is wrong... the value we have elsewhere is `SqlUnknownT "SQL_LONGVARCHAR"`.  I would double check this.

telser: Assuming that gets fixed I'd like to forward port this to the 0.9 branch too. I can do that work, but doesn't seem like it would be super difficult given what you've got here.

qxjit: @telser We have the appropriate `SqlType` value for the 0.9 development branch in several projects already. You could just grab one of those and then make sure whatever tests are in place in 0.9 get made for it too.

onslaughtq: @qxjit This works as written in schmods.  I went back and forth in my mind about using SqlVarCharT and SqlLongVarCharT, before landing on the former, because postgresql mostly says that Text and VarChar, without a length, are the same.  I'm not certain why you would use the value you mentioned.  
Also, as far as I can tell the actual SqlTypeId used for the SqlColDesc doesn't matter except for migrations.  As I have it, migrations from VarText n to UnboundedText and back work as expected.

qxjit: @onslaughtq If the `SqlTypeId` is not right, the symptom will be that orville will try to migrate the column to unbounded text *every time* migrations are run, even when there is no change in the code.

onslaughtq: @qxjit That is still not the behaviour I'm seeing, so I think this is fine. schmods migrated the columns to Text once, when I changed them. I changed some other columns to force new migrations, and the error column did not migrate a second time. I changed the error column back to VarText 255 and that worked as expected too.

qxjit: Upon further tandem investigation, it appears that the value being used on other projects is incorrect in at least one case.  We should move forward with this PR to `master`, but hold of on the `0.9_development` branch until we can determine whether there are projects that actually report a `SqlTypeId` other than `SqlVarCharT` or not.

PR #62, opened 2019-07-30
Doc fix and backport unbounded text

telser:


src/Database/Orville/Internal/Types.hs
```diff
@@ -37,6 +37,7 @@ data ColumnType
   = AutomaticId
   | ForeignId
   | Text Int
+  | UnboundedText
```
qxjit: I actually think this type is dead now. We should remove it rather than update it.

telser: Ah that makes sense

PR #65, opened 2019-08-12
Changes for working with Oracle's sql types.

telser:
- Adds field definitions for working with Oracle's NUMBER type which can be either Integers or Doubles.
- "Int" columns come back as bytestrings, so a, sadly hacky*, workaround is in place.
- Both "DATE" and "TIMESTAMP" in Oracle correspond to a LocalTime (use with localTimeField).
  Perhaps some aliases would be helpful? (eg dateField and/or timestampField pointing to localTimeField)
- Removes problematic escaping for now, though it _should_ be brought back when and where possible

Notablely some things do not seem to work yet:
- insertRecord fails with "Function sequence error", the statement preparation/execution is a candidate for further investigation
- migrateSchema causes Oracle to spin indefinitely.
  There are still some pg specifics to replace (listing constraints/indexes), but avoiding those still causes the CPU to sit at 100% with no end in sight
- selectFirst still has some sql incompatability

* This uses read to get the bytestrings back to ints :( suggestions on better ways to handle this welcome


orville-oracle/src/Database/Orville/Oracle/Internal/FromClause.hs
```diff
@@ -15,7 +15,7 @@ fromClauseRaw :: String -> FromClause
 fromClauseRaw = FromClause
 
 fromClauseTableName :: String -> FromClause
-fromClauseTableName name = fromClauseRaw ("FROM " ++ escapedName name)
+fromClauseTableName name = fromClauseRaw ("FROM " ++ name)
```
qxjit: Our poor escaping! 😢

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -248,22 +216,37 @@ localTimestamp =
...
+   really are integers -- ie "scale" of 0 with "precision" as input.
+-}
+integerNumber :: Int -> SqlType Integer
+integerNumber precision =
+  SqlType
+  { sqlTypeDDL = concat ["NUMBER(", show precision, ")"]
+  , sqlTypeReferenceDDL = Nothing
+  , sqlTypeNullable = False
+  , sqlTypeId = HDBC.SqlDoubleT -- Note that "NUMBER" is always decoded as Double by odbc
+  , sqlTypeSqlSize = Just (22*8) -- "NUMBER"s in Oracle can be up to 22 bytes!
```
qxjit: the size has nothing to do with the precision? Interesting!

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -328,6 +311,7 @@ int32FromSql sql =
   case sql of
     HDBC.SqlInt32 n -> Just n
     HDBC.SqlInteger n -> toBoundedInteger n
+    HDBC.SqlByteString n -> toBoundedInteger . read $ B8.unpack n
```
qxjit: We should at least use `readMaybe` here so that if the reading fails it gets surfaced as an error. attoparsec also has parsers for these, but I don’t know if we want to pull that in just for this.

This is one of those reason I really want to get off HDBC. Dealing this weirdness like this through the lens of `SqlValue` is quite vexing.

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -338,6 +322,7 @@ int64FromSql sql =
   case sql of
     HDBC.SqlInt64 n -> Just n
     HDBC.SqlInteger n -> toBoundedInteger n
+    HDBC.SqlByteString n -> toBoundedInteger . read $ B8.unpack n
```
qxjit: `readMaybe` again

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -396,6 +381,26 @@ localTimeFromSql sql =
...
 
+integerNumberToSql :: Integer -> HDBC.SqlValue
+integerNumberToSql = HDBC.SqlDouble . fromIntegral
+
+integerNumberFromSql :: HDBC.SqlValue -> Maybe Integer
+integerNumberFromSql sql =
+   -- Depending on how the column was created the "NUMBER" type can come back as a SqlDouble or a SqlByteString...
+  case sql of
+    HDBC.SqlDouble val -> doubleToInteger val
+    HDBC.SqlByteString bs -> doubleToInteger . read $ B8.unpack bs
```
qxjit: `readMaybe` 😉

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -396,6 +381,26 @@ localTimeFromSql sql =
...
+    _ -> Nothing
+
+doubleToInteger :: Double -> Maybe Integer
+doubleToInteger val =
+  let ceil = ceiling val
+  in
+    if ceil == floor val
```
qxjit: There’s really no better way to do check integrality?

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
telser: They even let you specify negative precision where 123456 will be stored as 123400 (with precision -2)!

orville-oracle/src/Database/Orville/Oracle/Internal/FromClause.hs
telser: I know! :( This is definitely done in the interest of speed.

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
telser: I was hoping to come up with something nicer than `read`/`readMaybe`, but perhaps attoparsec is a bit heavy handed. I've updated it to be `readMaybe` here and the others.

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
telser: I changed it to use `properFraction` which I thought would be roughly equivalent, but learned that `ceiling` and `floor` are both implemented in terms of `properFraction` so it is at least marginally faster/better.

PR #66, opened 2019-08-13
Updates to text fields for Oracle.

telser:
On text(read: CHAR(n)) columns, Oracle will right pad inserted values with spaces.
This manifested an issue when doing direct comparisons, such as in WHERE clauses, that
values did not compare in the expected behavior.

This changes the textField behavior to pad values on the way into the database
and strip whitespace on the way out.

Other minor change is moving an escaping helper to another module for reuse (when we finally add it back).
And an update to the sample project to reflect column names being capitalized in Oracle.

Other notes/Gotchas:
  - Orville does not have support for the Oracle types: VARCHAR2 or NVARCHAR2 so this has not been tested against those.
  - The behavior of the textField now means that data purposely ending with some number of spaces will not retain
    those spaces on the way out of the database, but potentially many more would have been added in the previous behavior.
  - Perhaps add "rawTextField" to allow users of Orville to handle the whitespace issue in application code, if needed?

Things still broken/needing attention:
  - The test suite needs to be completely reworked with instructions on using the available Oracle 12c docker image.
  - Several functions still do not work including:
    - `migrateSchema` (causes db thread to spin indefinitely)
    - `selectFirst` (uses incompatible sql)
    - `insertRecord` ("Function sequence error")

### DIFF
--- a/orville-oracle/sample-project/Example/Schema/Student.hs
+++ b/orville-oracle/sample-project/Example/Schema/Student.hs
@@ -38,17 +38,17 @@ majorTable =
 
 majorIdField :: O.FieldDefinition MajorId
 majorIdField =
-  O.int32Field "id" `O.withFlag` O.PrimaryKey `O.withConversion`
+  O.int32Field "ID" `O.withFlag` O.PrimaryKey `O.withConversion`
   O.convertSqlType majorIdInt MajorId
 
 majorNameField :: O.FieldDefinition MajorName
 majorNameField =
-  O.textField "name" 255 `O.withConversion`
+  O.textField "NAME" 255 `O.withConversion`
   O.convertSqlType majorNameText MajorName
 
 majorCollegeField :: O.FieldDefinition MajorCollege
 majorCollegeField =
-  O.textField "college" 255 `O.withConversion`
+  O.textField "COLLEGE" 255 `O.withConversion`
   O.convertSqlType collegeMajorToText textToCollegeMajor
 
 studentTable :: O.TableDefinition (Student StudentId) (Student ()) StudentId
--- a/orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
+++ b/orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
@@ -1,7 +1,7 @@
 module Database.Orville.Oracle.Internal.SqlType
   ( SqlType(..)
   , text
-  , varText
+  , autoPadAndStripText
   , integer
   , bigInteger
   , double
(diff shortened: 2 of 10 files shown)


orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -326,14 +326,22 @@ int64FromSql sql =
...
+  'textFromSql' strips leading and trailing whitespace from the resulting value to better reflect an
+  inverse of 'textToSql'
+-}
 textFromSql :: HDBC.SqlValue -> Maybe T.Text
 textFromSql sql =
   case sql of
-    HDBC.SqlByteString bytes -> Just $ Enc.decodeUtf8 bytes
-    HDBC.SqlString string -> Just $ T.pack string
+    HDBC.SqlByteString bytes -> Just . T.strip $ Enc.decodeUtf8 bytes
+    HDBC.SqlString string -> Just . T.strip $ T.pack string
```
qxjit: I'm not in favor of stripping the values in fixed width fields inside Orville itself. We should have a discussion.

telser: I am not comfortable with the way that shook out either. Glad to talk about it when you have a chance.

orville-oracle/sample-project/Example/Schema/Student.hs
```diff
@@ -38,17 +38,17 @@ majorTable =
...
 majorNameField :: O.FieldDefinition MajorName
 majorNameField =
-  O.textField "name" 255 `O.withConversion`
+  O.textField "NAME" 255 `O.withConversion`
   O.convertSqlType majorNameText MajorName
 
 majorCollegeField :: O.FieldDefinition MajorCollege
 majorCollegeField =
-  O.textField "college" 255 `O.withConversion`
+  O.textField "COLLEGE" 255 `O.withConversion`
```
qxjit: Is this because Oracle automatically upcases columns? Was this part of the migration problem?

orville-oracle/src/Database/Orville/Oracle/Internal/SqlType.hs
```diff
@@ -90,7 +90,7 @@ varText len =
     , sqlTypeNullable = False
     , sqlTypeId = HDBC.SqlVarCharT
     , sqlTypeSqlSize = Just len
-    , sqlTypeToSql = textToSql
+    , sqlTypeToSql = textToSql len
```
qxjit: Why would we pad a `VARCHAR`?

PR #75, opened 2020-03-10
[ch17460] Add support for partial indices

onslaughtq:
Two new functions have been added that allow for the
creation of a partial index or a unique partial index.

### DIFF
--- a/orville.cabal
+++ b/orville.cabal
@@ -2,7 +2,7 @@
 -- documentation, see http://haskell.org/cabal/users-guide/
 
 name:                orville
-version:             0.8.4.0
+version:             0.8.5.0
 synopsis:            ORM
 description:         ORM library for PostgreSQL
 license:             MIT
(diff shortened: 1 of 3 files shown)


onslaughtq:
@telser @qxjit This is failing to build using 7.10.3, seemingly unrelated to my change. Do I need to address this build issue, or is it known that the CI configuration for this old version may not successfully build?

qxjit:
@onslaughtq Without taking a lot of time to investigate, it looks like a newer version of the postgres library got release that may not be compatible with older versions of GHC?  In general, yes, I think we should only merge branches that pass the build. So in this situation, we have a few options:

1) fix the build
2) drop support for the old version of GHC on `master`.

Given that you're using the old GHC on Exfreight, I'd vote for the former so that we can confirm everything works on that GHC. While I don't doubt that that's the case, having a build to continue to confirm it in the future is good business.

If you need help fixing the build, I'm sure @telser or I can assist you.

PR #77, opened 2020-04-03
[ch18387] Add LIKE and ILIKE Where operators

onslaughtq:
This commit adds ~~ (LIKE) and ~~* (ILIKE) operators
for `where` conditions.

Note: The ILIKE operator is Postgres specific, and not
a SQL standard.

### DIFF
--- a/src/Database/Orville/Internal/Where.hs
+++ b/src/Database/Orville/Internal/Where.hs
@@ -39,13 +39,14 @@ instance QueryKeyable WhereCondition where
       => FieldDefinition -> a -> WhereCondition
 fieldDef .== a = BinOp "=" fieldDef (convert a)
 
-(.~~) :: Convertible a SqlValue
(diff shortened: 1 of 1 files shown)


src/Database/Orville/Internal/Where.hs
```diff
@@ -126,6 +135,18 @@ isNull fieldDef = IsNull fieldDef
 isNotNull :: FieldDefinition -> WhereCondition
 isNotNull fieldDef = IsNotNull fieldDef
 
+isLike :: FieldDefinition -> String -> WhereCondition
+isLike fieldDef pattern = fieldDef .~~ pattern
```
telser, line 139: Would it make more sense to define these the other way around? That the operator is a shortcut for the function rather than the function as a way to call the operator?

onslaughtq: Fixed

PR #79, opened 2020-07-22
Update MonadUnliftIO implementations for version 0.2.0

cmika:
Version 0.2.0 of unliftio-core removes `askUnliftIO` from the typeclass
in favor of `withRunInIO`. I've added a CPP check on the version of
unliftio-core to implement the right function for the right version.

I've also tested adding a restriction to unliftio-core for <0.2 which
also works. Please let me know if you would prefer that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -185,9 +185,9 @@ before_install:
 - |
   if [ `uname` = "Darwin" ]
   then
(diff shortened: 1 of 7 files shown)


liminalisht:
CI is still failing for stack lts 11

liminalisht:
yay

PR #82, opened 2020-07-23
Escape field names using NameForm

cmika:
This escapes field names in migrations, inserts and updates. This should
allow for the use of capital letters and reserved words as field names.

I chose to use NameForm over the exisiting `escapedName` functions for
consitency accross the places where field names are already being
escaped.


liminalisht:
+1 for tests. Just so I understand, will this affect existing "raw sql" queries in any way, e.g. queries where column names were not quoted?

cmika:
This should make it so all field names generated by orville will be quoted, which would change existing queries but not in a meaningful way. If you have raw sql this shouldn't affect it.

cmika:
I ran these tests on 0.9_development without my changes and they all failed with
```
Create table:                             FAIL
      Exception: There was an error migrating table entity.
      The error is:
      
       SqlError {seState = "42601", seNativeError = 7, seErrorMsg = "execute: PGRES_FATAL_ERROR: ERROR:  syntax error at or near \"order\"\nLINE 1: ...NOT NULL PRIMARY KEY, camelCase INTEGER NOT NULL, order INTE...\n                                                             ^\n"}\n
      
      Here are the developer comments regarding the table:
```

orville-postgresql/test/Migrations/MigrateTest.hs
```diff
@@ -0,0 +1,114 @@
...
+  testCase "Adding and Dropping columns is idempotent" $ do
+    run (TestDB.reset [O.Table tableDef])
+    assertMigrationIdempotent run [O.Table $ migrationEntityTableWithDroppedColumn columnName]
+    assertMigrationIdempotent run [O.Table tableDef]
+  where
+    tableDef :: O.TableDefinition
+                  (MigrationEntity Int32 MigrationEntityId)
+                  (MigrationEntity Int32 ())
+                  MigrationEntityId
+    tableDef = migrationEntityTable (O.int32Field columnName)
```
qxjit, line 75: I would maybe name this `tableDefWithColumn` and the other one `tableDefWithDroppedColumn` ?

Out of curiosity, did GHC required this type signature, or did you put it purely for communication purposes?

qxjit (approved):
:heart_eyes: Love these tests

orville-postgresql/test/Migrations/MigrateTest.hs
cmika: I didn't try it without the type signature, I thought it would be easier to understand when you could see the types

PR #83, opened 2020-09-03
Fixes Bug With Empty Conditionals

aaronallen8455:
Fixes an exception resulting from having an empty whereAnd/whereOr
condition.  It would print `()` in the sql which caused a postgres
syntax error.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Where.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Where.hs
@@ -119,22 +119,31 @@ whereConditionSql cond = internalWhereConditionSql Nothing cond
 
 internalWhereConditionSql ::
      Maybe (TableDefinition a b c) -> WhereCondition -> String
-internalWhereConditionSql (Just tableDef) (WhereConditionExpr expression) =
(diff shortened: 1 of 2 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Where.hs
```diff
@@ -119,22 +119,31 @@ whereConditionSql cond = internalWhereConditionSql Nothing cond
...
 internalWhereConditionSql ::
      Maybe (TableDefinition a b c) -> WhereCondition -> String
-internalWhereConditionSql (Just tableDef) (WhereConditionExpr expression) =
-  rawExprToSql . generateSql $ expression `qualified` (tableName tableDef)
-internalWhereConditionSql Nothing (WhereConditionExpr expression) =
-  rawExprToSql . generateSql $ expression
-internalWhereConditionSql tableDef (Or conds) =
-  List.intercalate " OR " condsSql
+internalWhereConditionSql mbTableDef whereCondition =
+  case whereCondition of
```
qxjit, line 123: :heart: case statement

qxjit (approved):

exodrifter (approved):

PR #84, opened 2020-09-15
Generate SQL for insertRecordMany

cmika:
Due to how `executeMany` is implmented the result of calling
`insertRecordMany` would be multiple calls to the database. This reworks
`insertRecordMany` so instead it generates a multi record insert
statement and only makes 1 call.


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Sql.hs
```diff
@@ -9,11 +9,23 @@ import qualified Data.List as List
 
 mkInsertClause :: String -> [String] -> String
 mkInsertClause tblName columnNames =
```
qxjit, line 11: Could `mkInsertClause` now just call `mkInsertManyClause` with a `recordCount` of 1?

qxjit (approved):
It might be nice to have a direct test of somewhere rather than relying on the usage in the popper tests.

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Sql.hs
cmika: I thought about that and forgot to change it lol

exodrifter (commented):
Do you also intend to fix this in [`Database.Orville.Oracle.Core.insertRecordMany`](https://github.com/flipstone/orville/blob/431f9aecca49dfcedb47c1d65a0c9bbc79faa191/orville-oracle/src/Database/Orville/Oracle/Core.hs#L308)?

cmika:
We don't have long term plans to support Oracle so I will not be making the change to Oracle @exodrifter

PR #85, opened 2020-09-16
Fix unboundedText's TypeId

cmika:
This corrects a bug where orville will attempt to migrate unboundedText
fields even though none is needed. This is because HDBC is reporting
SqlVarCharT as the TypeId.


qxjit (approved):
@cmika Can we add a test for this using new your new migration test stuff from the other PR?

exodrifter (approved):

PR #86, opened 2020-09-28
Begin basic libpq work.

telser:
This includes a minimal package.yaml/cabal file, and only a single stack-lts for the moment to keep things simple.

The code is the foundational connection and the pool interface is the same as previously with one exception:
  - the connection string is now a ByteString instead of String (though the underlying format is still the same)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
```diff
@@ -0,0 +1,83 @@
...
+        Nothing -> throwString "failed to get file descriptor"
+        Just fd' -> do
+          threadWaitFn fd'
+          poll conn
+    poll conn = do
+      pollStatus <- LibPQ.connectPoll conn
+      case pollStatus of
+        LibPQ.PollingFailed -> do
+          underlyingError <- (LibPQ.errorMessage conn)
+          throwString ("connection failure" <> show underlyingError)
```
aaronallen8455, line 62: perhaps add a ": " after "connection failure"

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
```diff
@@ -0,0 +1,83 @@
...
+  First note that async exceptions are masked.
+  `mask` though, only works for things that not interruptible <https://www.stackage.org/haddock/lts-16.15/base-4.13.0.0/Control-Exception.html#g:13>
+  From the previous link, `tryTakeMVar` is not interruptible, where `takeMVar` *is*.
+  So by using `tryTakeMVar` along with `mask`, we should be safe from async exceptions causing us to not finish an underlying connection.
+
+-}
+close :: Connection -> IO ()
+close (Connection handle') =
+  let underlyingFinish restore = do
+        underlyingConnection <- tryTakeMVar handle'
```
aaronallen8455, line 80: Doesn't this mean that if the connection is in use (MVar is empty) then it won't get closed? It seems like this _should_ block.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
```diff
@@ -0,0 +1,83 @@
...
+        LibPQ.PollingReading -> checkSocketAndThreadWait conn threadWaitRead
+        LibPQ.PollingWriting -> checkSocketAndThreadWait conn threadWaitWrite
+        LibPQ.PollingOk -> do
+          connectionHandle <- newMVar conn
+          pure (Connection connectionHandle)
+
+{-|
+  `close` has many subtleties to it.
+  First note that async exceptions are masked.
+  `mask` though, only works for things that not interruptible <https://www.stackage.org/haddock/lts-16.15/base-4.13.0.0/Control-Exception.html#g:13>
```
aaronallen8455, line 72: *are not

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
telser: It should never be empty. We don't expose the connection out and only control it here. If we blocked then an async exception could come in and cause us to not close.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
telser: Will fix

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
telser: They are masked, because we only call non-blocking functions on the inside.

orville-postgresql-libpq/package.yaml
```diff
@@ -0,0 +1,29 @@
...
+author: Flipstone Technology Partners
+maintainer: development@flipstone.com
+license: MIT
+git: git@github.com:flipstone/orville.git
+tested-with: GHC == 8.6.5
+extra-source-files:
+  - README.md
+ghc-options:
+  - -Wall
+  - -fno-warn-orphans
```
aaronallen8455, line 15: Do we really want this option to be set project wide?

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
aaronallen8455: sorry, just pointing out that it should read "that are not interruptible"

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
aaronallen8455: Ok, as long as the MVar isn't being used for anything else then that makes sense.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
```diff
@@ -0,0 +1,83 @@
...
+        Nothing -> throwString "failed to get file descriptor"
+        Just fd' -> do
+          threadWaitFn fd'
+          poll conn
+    poll conn = do
+      pollStatus <- LibPQ.connectPoll conn
+      case pollStatus of
+        LibPQ.PollingFailed -> do
+          underlyingError <- (LibPQ.errorMessage conn)
+          throwString ("connection failure: " <> show underlyingError)
```
qxjit, line 61: I'd prefer to throw our own exception type rather than using `throwString`

qxjit (changes requested):
Looks great -- I'd just rather throw our own type rather than `throwString`

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
telser: I've updated with a custom type, though it is probably going to need some iteration as we expand things.

qxjit (approved):
@telser :heart:

PR #87, opened 2020-09-29
RFC: Initial port of sqltype

telser:
I would really love feedback on this before getting much further. As it stands this is a pretty basic translation of `SqlType` but there are a few things to think about for driving forward:

1. Do we want the functions for conversion to take/return `ByteString`?
   The underlying LibPQ stuff all works on `ByteString`s so at some point we'll have to be there but we could..
   a) Use a type alias to at least give some hint to users that this is the PostgreSQL representation of things
   b) Use a newtype wrapper to force _some_ thought around working with these.
   c) Use the plain `ByteString` which is arguably the least amount of work when creating a `SqlType`
2. Do we want the reading of a value from PostgreSQL to return `Maybe a`?
    I did it this way because that is how Orville operates with `HDBC` today, but maybe it would be better to have some way to report errors back out?
3. This code uses `attoparsec` to do the ByteString to Int32 conversion (which we could extend/reuse easily to other Integral types). Note that end users would not be forced to directly interact with `attoparsec` and when writing custom `SqlType`s could use whatever they wanted. Still, how do we feel about this as a dependency? Any suggestions on using something else?

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
@@ -1,0 +1,31 @@
+{-|
+Module    : Database.Orville.PostgreSQL.Raw
+Copyright : Flipstone Technology Partners 2020
+License   : MIT
+-}
+
+module Database.Orville.PostgreSQL
+  ( createConnectionPool
+  , SqlType ( SqlType
+            , sqlTypeDDL
+            , sqlTypeReferenceDDL
+            , sqlTypeNullable
+            , sqlTypeId
+            , sqlTypeSqlSize
+            , sqlTypeToSql
+            , sqlTypeFromSql
+            )
+  )
+where
+
+import Database.Orville.PostgreSQL.Connection (createConnectionPool)
+import Database.Orville.PostgreSQL.Internal.SqlType (SqlType ( SqlType
+                                                             , sqlTypeDDL
+                                                             , sqlTypeReferenceDDL
+                                                             , sqlTypeNullable
+                                                             , sqlTypeId
+                                                             , sqlTypeSqlSize
+                                                             , sqlTypeToSql
+                                                             , sqlTypeFromSql
+                                                             )
+                                                    )
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
@@ -15,6 +15,7 @@ module Database.Orville.PostgreSQL
             , sqlTypeToSql
             , sqlTypeFromSql
             )
+  , integer
   )
 where
 
@@ -28,4 +29,5 @@ import Database.Orville.PostgreSQL.Internal.SqlType (SqlType ( SqlType
                                                              , sqlTypeToSql
                                                              , sqlTypeFromSql
                                                              )
+                                                    , integer
                                                     )
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
@@ -16,8 +16,7 @@ module Database.Orville.PostgreSQL
             , sqlTypeFromSql
             )
   , integer
-  )
-where
+  ) where
 
 import Database.Orville.PostgreSQL.Connection (createConnectionPool)
 import Database.Orville.PostgreSQL.Internal.SqlType (SqlType ( SqlType
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
@@ -14,10 +14,16 @@ module Database.Orville.PostgreSQL.Internal.SqlType
            , sqlTypeToSql
            , sqlTypeFromSql
            )
+  , integer
   )
 where
 
+import Data.Attoparsec.ByteString (parseOnly)
+import Data.Attoparsec.ByteString.Char8 (signed, decimal)
 import Data.ByteString (ByteString)
+import Data.ByteString.Builder (int32Dec, toLazyByteString)
+import Data.ByteString.Lazy (toStrict)
+import Data.Int (Int32)
 import qualified Database.PostgreSQL.LibPQ as LibPQ
 
 {-|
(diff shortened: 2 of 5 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -0,0 +1,81 @@
...
+    , sqlTypeId = LibPQ.Oid 23
+    , sqlTypeSqlSize = Just 4
+    , sqlTypeToSql = int32ToBS
+    , sqlTypeFromSql = int32FromBS
+    }
+
+int32ToBS :: Int32 -> ByteString
+int32ToBS = toStrict . toLazyByteString . int32Dec
+
+int32FromBS :: ByteString -> Maybe Int32
```
liminalisht, line 78: I'm very much in favor of using Either here and giving some indication of the reason the parse failed

telser: Yeah, I like that idea too. There could be some conflating of "I know I cannot convert this" and "I thought I could, but failed". In which case I see three paths:
- Live with that and return `Either String a` where the `String` is an explanation of what went wrong
- Have a custom type for the `Left` side to differentiate those two cases that is along the lines of
```haskell
data ConversionFailure = WillNotConvert | ConversionAttemptFailed String
```
- Use a custom type for the whole thing that includes success and the various failure modes.

I like the clarity of the 2nd/3rd options, but want to weigh carefully the cost of that for users.

telser: Note that this particular example bit could be extended to use the `WillNotConvert` bit if given an empty input.. Trying to capture the meaning of the `Maybe` as it is today, i.e. Nothing if conversion is not possible and we know that.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
```diff
@@ -0,0 +1,33 @@
...
+            , sqlTypeReferenceDDL
+            , sqlTypeNullable
+            , sqlTypeId
+            , sqlTypeSqlSize
+            , sqlTypeToSql
+            , sqlTypeFromSql
+            )
+  , integer
+  )
+where
```
qxjit, line 20: Where next to `)` please

orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
```diff
@@ -0,0 +1,33 @@
...
+                                                             , sqlTypeDDL
+                                                             , sqlTypeReferenceDDL
+                                                             , sqlTypeNullable
+                                                             , sqlTypeId
+                                                             , sqlTypeSqlSize
+                                                             , sqlTypeToSql
+                                                             , sqlTypeFromSql
+                                                             )
+                                                    , integer
+                                                    )
```
qxjit, line 33: I would really prefer qualified imports to many-lined imports

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -0,0 +1,81 @@
...
+int32FromBS bs = case parseOnly (signed decimal) bs of
+  Left _ -> Nothing
+  Right i -> Just i
```
qxjit, line 81: Integer values come back from `libpq` as values that we parse as strings? This surprises me a bit. I would have thought they would come back in more directly encoded as a binary number

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
qxjit: It's not clear to me how separate the 2 failure cases for this example matters.

I'd love to be able to get some error messaging in here as well, but one word of caution -- it's very tempting to put the value you're attempting to decode in the error message. This could potentially be disastrous because it's very easy for errors to end up being sent to places where sensitive data should not go, and these generic low-level decodes have no knowledge about the sensitivity of the data they're decoding. That's one reason we use a `Maybe` today because it takes away the option of including potentially sensitive data in the error message.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -0,0 +1,81 @@
...
+  , sqlTypeNullable :: Bool
+    -- ^ Indicates whether columns should be marked NULL or NOT NULL in the
+    -- database schema. If this is 'True', then 'sqlTypeFromSql' should
+    -- provide a handling of 'SqlNull' that returns an 'a', not 'Nothing'.
+  , sqlTypeId :: LibPQ.Oid
+  , sqlTypeSqlSize :: Maybe Int
+  , sqlTypeToSql :: a -> ByteString
+    -- ^ A function for converting Haskell values of this type into values to
+    -- be stored in the database.
+  , sqlTypeFromSql :: ByteString -> Maybe a
```
qxjit, line 52: I imagine we may eventually want something richer than just `ByteString` here, but I can't really say what it is at this early stage.

qxjit (approved):
All in all there's nothing particular controversial here, I don't think. I'm sure we'll evolve `SqlType` as we get more interactions with `libpq` under our belt, but we don't have the knowledge to do so a priori. Starting with a direct port of the current Orville `SqlType` is a fine starting point.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
telser: But this is the re-exporter, we don't have much choice unless we always re-export everything with the `as` but no `qualified` or have an open import. Both increase the chance of accidentally add something to the public interface, though not to the same degree. I would prefer us to be very explicit/intentional about what is exposed.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
telser: Done

orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
qxjit: I agree we should be explicit. I didn't notice that this was a new file that was only for re-export.

Maybe we don't have to be explicit with every single field of records where we export the constructor though? If we're exporting the constructor not exporting an individual field doesn't buy you anything. Instead explicitly exporting each field opens up the possibility that you add a field and forget to export it.

PR #89, opened 2020-12-07
Add streaming pagination

aaronallen8455:
The `conduitSelect` function does not work as intended - the entire
query result is loaded into memory before the stream can begin. This is
because if single row mode is not used, postgres collects all the
results before returning to the application (see
https://www.postgresql.org/docs/9.2/libpq-single-row-mode.html).

It would be possible to implement the correct streaming functionality
using low level libpq functions, but hdbc-postgres would need to provide
an interface for them which it does not.

This PR implements the next best thing, which is page based streaming.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Conduit.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Conduit.hs
@@ -7,6 +7,7 @@ License   : MIT
 
 module Database.Orville.PostgreSQL.Conduit
   ( selectConduit
+  , streamPages
   ) where
 
 {-
@@ -26,6 +27,7 @@ import Conduit
   , allocateAcquire
   , mkAcquire
   , mkAcquireType
+  , yieldMany
   )
 
 import Control.Monad (void)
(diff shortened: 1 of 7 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Conduit.hs
```diff
@@ -39,18 +41,24 @@ import Database.HDBC hiding (withTransaction)
...
+   CSV). This is useful if you want to be able to process many rows one by one.
+   You can aggregate the results however you require as part of the conduit
+   processing and then use 'runConduit' (or 'runConduitRes') from the conduit
+   library to execute the processing pipeline. Alternatively, your web server
+   ('wai', 'servant', etc) may provide support for converting a conduit into a
+   streaming HTTP response.
+
+   Beware: this function must load all the results into memory before streaming
+   can begin. For why, see https://www.postgresql.org/docs/9.2/libpq-single-row-mode.html.
+   If memory use is a concern, try 'streamPages' instead.
```
qxjit, line 61: I imagine we'll fix this with the switch to `libpq`

qxjit (approved):
I'm happy to have this merged -- it might see significant change with the port to `libpq` though.

orville-postgresql/src/Database/Orville/PostgreSQL/Conduit.hs
aaronallen8455: Agreed

PR #90, opened 2020-12-07
Generalize Transactions Over an OrvilleEnv

aaronallen8455:
Exposes a new function, `withTransactionOverEnv` which can be used to
layer transactions over multiple databases - something that WIMS benefits
greatly from.


qxjit:
I'd rather not see this go in -- exposing operations that take the `OrvilleEnv` directly isn't something I'd imagined ever doing, and holding transactions open against multiple databases generally something to be avoided anyway. I'm fairly sure you could do it without changes to Orville if you really want to, it just may involve a little bit of `Reader` context juggling.

aaronallen8455:
It's not possible without changing orville due to `OrvilleEnv` being abstract and the functions operating on it are bracketed so that connections, transaction state, etc cannot be juggled (rightly so).

PR #91, opened 2020-12-17
This brings the libpq `SqlType` module up to parity with the current version in terms of types supported.

telser:
Oids were manually verified first in psql

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
@@ -289,10 +289,10 @@ nullableType :: SqlType a -> SqlType (Maybe a)
 nullableType sqlType =
   sqlType
     { sqlTypeNullable = True
-    , sqlTypeToSql = maybe (toStrict . toLazyByteString $ stringUtf8 "null") (sqlTypeToSql sqlType)
+    , sqlTypeToSql = maybe nullBS (sqlTypeToSql sqlType)
     , sqlTypeFromSql =
         \sql ->
-          if sql == (toStrict . toLazyByteString $ stringUtf8 "null") then
+          if sql == nullBS then
             Just Nothing
           else
             fmap Just (sqlTypeFromSql sqlType sql)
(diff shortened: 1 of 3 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -71,10 +98,301 @@ integer =
...
+   represent NULL values when converting to and from sql.
+  -}
+nullableType :: SqlType a -> SqlType (Maybe a)
+nullableType sqlType =
+  sqlType
+    { sqlTypeNullable = True
+    , sqlTypeToSql = maybe (toStrict . toLazyByteString $ stringUtf8 "null") (sqlTypeToSql sqlType)
+    , sqlTypeFromSql =
+        \sql ->
+          if sql == (toStrict . toLazyByteString $ stringUtf8 "null") then
```
qxjit, line 295: I would extract `toStrict . toLazyByteString $ stringUtf8 "null"` to a top level constant

qxjit (approved):
Looks good. One small request and one question in general

Request: Can we standardize around functions always having a newline after `=` in their definitions in this codebase?

Question: Are we close to being able to write tests focused tests for these column types?

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
telser: > Looks good. One small request and one question in general
> 
> Request: Can we standardize around functions always having a newline after `=` in their definitions in this codebase?
> 
I'm fine with that, but unless we have some way to enforce it.. Do we want to keep using hindent? I can bring the config for that over and start getting that going.

> Question: Are we close to being able to write tests focused tests for these column types?

Not exactly, the best way is probably to get a minimal select up with rawsql which I think is two stories away.

telser:
> Looks good. One small request and one question in general
> 
> Request: Can we standardize around functions always having a newline after `=` in their definitions in this codebase?
> 
I'm fine with that, but unless we have some way to enforce it.. Do we want to keep using hindent? I can bring the config for that over and start getting that going.

> Question: Are we close to being able to write tests focused tests for these column types?

Not exactly, the best way is probably to get a minimal select up with rawsql which I think is two stories away.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -71,10 +98,322 @@ integer =
...
+    txt <- textFromBS bs
+    Time.parseTimeM False Time.defaultTimeLocale (Time.iso8601DateFormat Nothing) (unpack txt)
+
+utcTimeToBS :: Time.UTCTime -> ByteString
+utcTimeToBS =
+  B8.pack . Time.formatTime Time.defaultTimeLocale (Time.iso8601DateFormat Nothing)
+
+utcTimeFromBS :: ByteString -> Maybe Time.UTCTime
+utcTimeFromBS bs =
+  do
```
qxjit, line 412: :laughing: TBH, I didn't mean before `do` (though I see why did that given my comment), but I don't actually hate it

qxjit:
@telser Oh, I forgot about `hindent`.  I've never been fully satisfied with its formatting though, so I would say let's drop it at this point since we're essentially rewriting.  If we want to give a different formatter a try I'm game though (`ormulo` maybe?)

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -71,10 +98,322 @@ integer =
...
+    txt <- textFromBS bs
+    Time.parseTimeM False Time.defaultTimeLocale (Time.iso8601DateFormat Nothing) (unpack txt)
+
+-- | NULL as a bytestring
+nullBS :: ByteString
+nullBS =
+  toStrict . toLazyByteString $ stringUtf8 "null"
```
qxjit, line 419: There's something very satisfying about naming something `nullBS`

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
telser: I'm never going to be 100% happy with the formatting so I'll do whatever. I actually dislike the newline before `do` and in case of the single line functions, but :shrug:

qxjit: I'm happy with `= do` and having the body on the next line.

I hear you about 1 line functions, but I've found that the consistency has grown on me over time. That said, I'm not going to get fussy about 1 liners either.  `case` and `if` statements directly after `=` particularly irk me though.

I'm fine with us removing the `hindent` stuff and going without standard formatting as well.

PR #92, opened 2020-12-21
Show the SqlType stuff in action - albeit rough action

telser:
A bunch of stuff going on here:
- Fixes an issue (and adds some comment) around the UTCTime decoding
- Adds a very "Raw" execution interface
- Includes a naive way to decode rows with the intent of showing the SQLType stuff working
- Has a manual 'visual test' set that can be loaded at the repl to manually show the various SQLType decoding do in fact work

The test should be turned into a real test suite with better values, docker, and CI. But that's being left for future PRs.

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
@@ -15,7 +15,29 @@ module Database.Orville.PostgreSQL
             , sqlTypeToSql
             , sqlTypeFromSql
             )
+    -- numeric types
   , integer
+  , serial
+  , bigInteger
+  , bigserial
+  , double
+
+    -- textual-ish types
+  , boolean
+  , unboundedText
+  , fixedText
+  , boundedText
+  , textSearchVector
+
+    -- date types
+  , date
+  , timestamp
+
+    -- type conversions
+  , nullableType
+  , foreignRefType
+  , convertSqlType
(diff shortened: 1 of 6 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL.hs
```diff
@@ -28,5 +50,27 @@ import Database.Orville.PostgreSQL.Internal.SqlType (SqlType ( SqlType
...
+
+                                                    -- date types
+                                                    , date
+                                                    , timestamp
+
+                                                    -- type conversions
+                                                    , nullableType
+                                                    , foreignRefType
+                                                    , convertSqlType
+                                                    , maybeConvertSqlType
```
qxjit, line 75: How would you feel about these being qualified imports and then using qualifications on the values in the exports to avoid listing them twice?

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -179,11 +178,11 @@ boolean =
     }
 
 {-|
-  'text' defines a fixed length text field type. This corresponds to a
+  'unboundedText' defines a fixed length text field type. This corresponds to a
```
qxjit, line 181: Oops, this probably shouldn't say "fixed length".

qxjit (approved):
:heart_eyes:

PR #93, opened 2021-01-04
Beginnings of being able to generate SQL

telser:
Little manual test is updated to use this to generate the select and inserts


qxjit (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
```diff
@@ -0,0 +1,63 @@
...
+
+data InsertExpr =
+  InsertExpr
+    { target :: T.Text
+    , rowValues :: [T.Text]
+    }
+
+insertExprToSql :: InsertExpr -> B8.ByteString
+insertExprToSql insertExpr =
+  TextEnc.encodeUtf8 $
```
aaronallen8455, line 56: Perhaps use `ByteString` throughout unless there's a specific reason for `Text`.

aaronallen8455 (approved):

PR #94, opened 2021-02-05
Removes `streamPages` if Conduit is less than 1.3

jbrechtel:
The previous incarnation was preventing compilation on conduit < 1.3 as
parts of Pagination weren't imported then so this just removes that
function completely.


aaronallen8455:
> parts of Pagination weren't imported 

I'm a little confused as to why this is the case.

aaronallen8455 (approved):

qxjit (approved):

PR #97, opened 2021-02-17
Adds a `withCachedConnection` which will checkout a connection before running an action.

telser:
This is similar to `withTransaction`, except this does *not* do any of the transaction setup.

Note that all the heavy lifting is actually done by `withConnectionEnv` and this is just a thin wrapper
around that function which we expose out of the library.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Raw.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Raw.hs
@@ -45,7 +45,10 @@ updateSql sql values =
 startTransaction :: ConnectionEnv conn -> ConnectionEnv conn
 startTransaction c = c {ormTransactionOpen = True}
 
--- | Runs an action with a cached connection
+-- | Runs an action with a cached connection.
+--   Without using this, or wrapping calls in a transaction using `withTransaction`, successive
+--   calls to functions like `insertRecord` and `updateRecord` are *not* guaranteed to occur on the
+--   same connection.
 withCachedConnection :: MonadOrville conn m => m a -> m a
 withCachedConnection action =
   withConnectionEnv (const action)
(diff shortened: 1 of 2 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Raw.hs
```diff
@@ -44,6 +45,11 @@ updateSql sql values =
 startTransaction :: ConnectionEnv conn -> ConnectionEnv conn
 startTransaction c = c {ormTransactionOpen = True}
 
+-- | Runs an action with a cached connection
+withCachedConnection :: MonadOrville conn m => m a -> m a
+withCachedConnection action =
+  withConnectionEnv (const action)
```
telser, line 51: The biggest thing here was we already had this ability just in a module that isn't then exposed and it took a `ConnectionEnv`, but in this case we don't need to do anything special with the connection just make sure we already have one.

orville-postgresql/src/Database/Orville/PostgreSQL/Raw.hs
```diff
@@ -44,6 +45,11 @@ updateSql sql values =
 startTransaction :: ConnectionEnv conn -> ConnectionEnv conn
 startTransaction c = c {ormTransactionOpen = True}
 
+-- | Runs an action with a cached connection
```
aaronallen8455, line 48: perhaps explain why someone might want to do this?

aaronallen8455 (approved):

orville-postgresql/src/Database/Orville/PostgreSQL/Raw.hs
telser: That's a great point. I'm not sure how much detail we want to get into with the docstring here. 

Avoiding an exception killing the connection is the motivator, but also anywhere that you want to grab a connection do potentially many things. Wording this well without making it seem like the only use case is the exception thing feels tricky to me at the moment. Any suggestions?

qxjit: I would start with explaining that _without_ calling this, there's no guarantee that individual operations like `insertRecord` happen on the same connection -- that each one checks out a connection from the pool. I wouldn't hold up this PR looking for perfect docs here though, just get in something with a bit more detail for us to improve later before release.

qxjit (approved):

PR #101, opened 2021-03-20
Fixes `maybeMapper` breaking on already nullable fields

qxjit:
This changes `nullableType` to be aware of whether it is wrapping itself
around a type that is already nullable. In that case it uses the inner
type's 'NULL' decode rather than returning a `Nothing`. Essentially this
preserves as much information as possible about the value at the time of
decoding.

Compare with #102 for an alternative, less subtle fix.


jbrechtel (commented):
I don't hate this TBH

jbrechtel (approved):

qxjit:
Closing in favor of #102

PR #102, opened 2021-03-20
Tracks field nullability at the type level in `FieldDefinition`

qxjit:
This fixes `maybeMapper` being used on already nullable fields by
tracking nullability as a new type parameter on `FieldDefinition` to
prevent `nullableField` ever being called on an already nullable field.
This replaces the `sqlTypeNullable` field on `SqlType` completely.

I opted to put the type parameter on `FieldDefinition` rather than
`SqlType` for several reasons

1) It has to be there anyway for enforcement on `nullableField` to work

2) Avoiding `SqlType` minimizes the number of types impacted by the
extra parameter.

3) It seems more accurate to model _fields_ as nullable rathar than
_types_. In a (probably too quick) perusal of the PostgreSQL docs on
types I couldn't actually find a reference to nullability.

Constructorless tag types `Nullable` and `NotNull` are used to fill the
parameter with concrete types as appropriate. Each implements the
`Nullability` class, which provides the `checkNullability` function for
evaluating the nullability at run time.

With `nullableField` barred from being called on `Nullable` fields,
`maybeMapper` now required a new function to handle cases where it is
used over such fields. I added `assymmetricNullableField` to satisfy its
need.

Compare with #101 for a fix that does not involve adding type-level tracking

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
@@ -95,7 +95,7 @@ mkTableDefinition (TableParams {..}) =
     }
 
 data RelationalMap a b where
-  RM_Field :: Nullability nullability => FieldDefinition nullability a -> RelationalMap a a
+  RM_Field :: FieldDefinition nullability a -> RelationalMap a a
   RM_Nest :: (a -> b) -> RelationalMap b c -> RelationalMap a c
   RM_Pure :: b -> RelationalMap a b
   RM_Apply
@@ -119,7 +119,7 @@ instance Profunctor RelationalMap where
 mapAttr :: (a -> b) -> RelationalMap b c -> RelationalMap a c
 mapAttr = RM_Nest
 
-mapField :: Nullability nullability => FieldDefinition nullability a -> RelationalMap a a
+mapField :: FieldDefinition nullability a -> RelationalMap a a
 mapField = RM_Field
 
 partialMap :: RelationalMap a (Either String a) -> RelationalMap a a
@@ -128,10 +128,10 @@ partialMap = RM_Partial
 readOnlyMap :: RelationalMap a b -> RelationalMap c b
 readOnlyMap = RM_ReadOnly
 
-attrField :: Nullability nullability => (a -> b) -> FieldDefinition nullability b -> RelationalMap a b
+attrField :: (a -> b) -> FieldDefinition nullability b -> RelationalMap a b
 attrField get = mapAttr get . mapField
 
-readOnlyField :: Nullability nullability => FieldDefinition nullability a -> RelationalMap b a
+readOnlyField :: FieldDefinition nullability a -> RelationalMap b a
 readOnlyField = readOnlyMap . mapField
 
 prefixMap :: String -> RelationalMap a b -> RelationalMap a b
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
@@ -162,9 +162,9 @@ maybeMapper
         NullableField nullField ->
           -- When the underlying field is already nullable we need to make sure
           -- that 'NULL' is decoded to a 'Just'. Otherwise when the field is
-          -- 'NULL' it cause the entire 'RelationalMap' to resolve to a
+          -- 'NULL' it causes the entire 'RelationalMap' to resolve to a
           -- 'Nothing' as if _all_ fields were 'NULL', even if they were not.
-          RM_Field (assymmetricNullableField nullField)
+          RM_Field (asymmetricNullableField nullField)
 
     go (RM_Pure a) = RM_Pure (pure a)
     go (RM_Apply rmF rmA) = RM_Apply (fmap (<*>) $ go rmF) (go rmA)
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
@@ -49,7 +49,7 @@ data TableParams readEntity writeEntity key = TableParams
   , tblSafeToDelete :: [String]
       -- ^ A list of any columns that may be deleted from the table by Orville.
       -- (Orville will never delete a column without being told it is safe)
-  , tblPrimaryKey :: FieldDefinition key
+  , tblPrimaryKey :: FieldDefinition NotNull key
       -- ^ A function to set the key on the entity
   , tblGetKey :: readEntity -> key
       -- ^ A function to get the key on the entity
@@ -95,7 +95,7 @@ mkTableDefinition (TableParams {..}) =
     }
 
 data RelationalMap a b where
-  RM_Field :: FieldDefinition a -> RelationalMap a a
+  RM_Field :: Nullability nullability => FieldDefinition nullability a -> RelationalMap a a
   RM_Nest :: (a -> b) -> RelationalMap b c -> RelationalMap a c
   RM_Pure :: b -> RelationalMap a b
   RM_Apply
@@ -119,7 +119,7 @@ instance Profunctor RelationalMap where
 mapAttr :: (a -> b) -> RelationalMap b c -> RelationalMap a c
 mapAttr = RM_Nest
 
-mapField :: FieldDefinition a -> RelationalMap a a
+mapField :: Nullability nullability => FieldDefinition nullability a -> RelationalMap a a
 mapField = RM_Field
 
 partialMap :: RelationalMap a (Either String a) -> RelationalMap a a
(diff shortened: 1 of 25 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/RelationalMap.hs
```diff
@@ -154,7 +154,18 @@ maybeMapper
...
-    go (RM_Field f) = RM_Field (nullableField f)
+    go (RM_Field f) =
+      case checkNullability f of
+        NotNullField notNullField ->
+          RM_Field (nullableField notNullField)
+
+        NullableField nullField ->
+          -- When the underlying field is already nullable we need to make sure
+          -- that 'NULL' is decoded to a 'Just'. Otherwise when the field is
+          -- 'NULL' it cause the entire 'RelationalMap' to resolve to a
```
telser, line 165: cause > causes?

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Types.hs
```diff
@@ -183,7 +200,7 @@ data TableDefinition readEntity writeEntity key = TableDefinition
   , tableSafeToDelete :: [String]
       -- ^ A list of any columns that may be deleted from the table by Orville.
       -- (Orville will never delete a column without being told it is safe)
-  , tablePrimaryKey :: FieldDefinition key
+  , tablePrimaryKey :: FieldDefinition NotNull key
```
jbrechtel, line 203: This is nice

jbrechtel (commented):
Tagging the type with this seems more transparent than the alternative and the primary key `FieldDefinition` builder fn gets a nice win out of this.

Just to confirm I'm not missing something, though - this *doesn't* prevent mixing `Nullable` and `NOT NULL` (or vice versa) - does it?

aaronallen8455 (commented):
There's a few things I don't like about this
- The `FieldDefinition` type has some clunkiness in that `Nullability` and the `Maybe`ness of the field are disjoint, i.e. the type permits `FieldDefinition NotNull (Maybe ...)` and `FieldDefinition Nullable Int`. I can see this being a point of confusion.
- The type level tag is only really used in a few places but leaks across much of the codebase. I'm not convinced the value it brings justifies this. Fwiw, you could accomplish the same thing with
```
type family FieldNotMaybe a where
  FieldNotMaybe (Maybe a) = TypeError (Text "Field is already nullable.")
  FieldNotMaybe a = ()

nullableField :: FieldNotMaybe a => FieldDefinition a -> FieldDefinition (Maybe a)
```
Which allows for a more explicit type error and doesn't leak complexity.

qxjit:
@aaronallen8455 Nothing says that a `NULL` value from the database always has to be handled as maybe from in `Haskell`. Having the two concepts be separate but related allows for the potential of mapping `NULL` values in some other way.  The type family you suggest assumes that the `Maybe`-ness indicates nullability, but that may not actually be the case. Maybe `Nothing` is getting serialized to the DB as an empty string for a not-null text column rather than a nullable column.

That said, I agree I don't love how far the nullability parameter needs to spread for this. It _does_ have the nice upside that @jbrechtel mentioned about barring NULL columns from the primary key field, however.

qxjit:
@jbrechtel If by "mixing nullable and not-nulll" you mean using both kinds of field definitions in the same `RelationalMap` then you're correct. This still allows them to be mixed because the `RM_Field` constructor is existentially quantified on the `nullability` variable, thereby swallowing it to build a `RelationalMap a a` without mentioning `nullability`.

aaronallen8455:
> @aaronallen8455 Nothing says that a `NULL` value from the database always has to be handled as `Maybe` in Haskell.
No, but our types definitely suggest it. It might be better to have `Nullable a` wrapper if conflating `Maybe` and nullability is undesirable.

qxjit:
@aaronallen8455 I think it's important to separate the part of the `FieldDefinition` that indicates "this field is nullable in the database" from the part that defines the Haskell type to represent the value. It's more declarative as to what the actual DB schema will be, rather than relying on a convention that the user needs to be aware of.  `FieldDefinition Null Int` is not inherently invalid, anyhow. It would just have to be constructed in such a way that null values are mapped to `Int` values when read from the database. That could be done today by starting with `FieldDefinition Null (Maybe Int)` and then applying the appropriate conversion to map `Maybe Int` into `Int`. We still want to stop `nullableField` from being called again on that sort of field, since it's already nullable even though `Int` itself does not indicate that.

aaronallen8455:
@qxjit I guess I had forgotten that orville is totally open-ended with regards to conversion between representations. In this light, nullability truly is just a tag that we want to propagate through some arbitrary transformations. I think I'd feel better about it being a concrete part of the haskell representation but that's not entirely possible in this context.

qxjit:
@aaronallen8455 Agreed -- The `Nullability` typeclass made me nervous from the beginning. I pushed a revision that replaces it with a `Nullability` GADT value that still surfaces the nullability to the type level. I hope that helps tie things together between the `FieldDefinition` representation and its type. It also causes less pollution in other types because the `Nullability` is no longer a constraint that needs to be tracked.

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Types.hs
```diff
@@ -62,18 +62,35 @@ instance ColumnDefault Bool where
...
   NotNull  :: Nullability NotNull
 
+-- | A 'NullabilityCheck' is returned by the 'checkNullability' function, which
+-- can be used when a function works on both 'Nullable' and 'NotNull' functions
+-- but needs to deal with each type of field separately. It adds wrapper
+-- constructors around the 'FieldDefinition' that you can pattern match on to
+-- then work with a concrete 'Nullable' or 'NotNull' field.
 data NullabilityCheck a
   = NullableField (FieldDefinition Nullable a)
   | NotNullField (FieldDefinition NotNull a)
```
aaronallen8455: I'm not sure if this type is really necessary. You should be able to just pattern matching on the GADT contained in a `FieldDefinition` to instantiate it's type variable.

orville-postgresql/test/OptionalMap/Entity.hs
```diff
@@ -0,0 +1,54 @@
...
+badComplexFieldMap =
+  ComplexField <$> O.attrField complexFieldPart1 complexFieldPart1Field
+               <*> O.attrField complexFieldPart2 (O.nullableField complexFieldPart2Field)
+
+complexFieldPart1Field :: O.FieldDefinition O.NotNull Int32
+complexFieldPart1Field = O.int32Field "part_1"
+
+complexFieldPart2Field :: O.FieldDefinition O.NotNull T.Text
+complexFieldPart2Field =
+  O.withConversion (O.textField "part_2" 255) $ O.maybeConvertSqlType id Just
```
aaronallen8455, line 54: What's the purpose of this conversion?

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Types.hs
```diff
@@ -62,16 +62,57 @@ instance ColumnDefault Bool where
...
+-- then work with a concrete 'Nullable' or 'NotNull' field.
+data NullabilityCheck a
+  = NullableField (FieldDefinition Nullable a)
+  | NotNullField (FieldDefinition NotNull a)
+
+-- | Resolves the 'nullablity' of a field to a concrete type based on its
+-- 'fieldNullability'. You can do this directly by pattern matching on the
+-- value of 'fieldNullability' if you have the GADTs extension turned on, but
+-- this function will do that for you so you don't need to turn GADTs on.
+checkNullability :: FieldDefinition nullability a -> NullabilityCheck a
```
jbrechtel, line 97: Oh wow, I really appreciate the GADT hiding here

jbrechtel (approved):
:heart:

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Types.hs
```diff
@@ -62,16 +62,57 @@ instance ColumnDefault Bool where
...
+  case fieldNullability field of
+    Nullable -> NullableField field
+    NotNull  -> NotNullField field
+
+data FieldDefinition nullability a =
+  FieldDefinition
+    { fieldName        :: String
+    , fieldType        :: SqlType a
+    , fieldFlags       :: [ColumnFlag]
+    , fieldNullability :: Nullability nullability
```
aaronallen8455, line 108: Since this constructor is fully exposed, the `nullability` evidence is rather fragile. I think it should be abstract if possible.

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Types.hs
```diff
@@ -62,16 +62,57 @@ instance ColumnDefault Bool where
...
+-- 'fieldNullability'. You can do this directly by pattern matching on the
+-- value of 'fieldNullability' if you have the GADTs extension turned on, but
+-- this function will do that for you so you don't need to turn GADTs on.
+checkNullability :: FieldDefinition nullability a -> NullabilityCheck a
+checkNullability field =
+  case fieldNullability field of
+    Nullable -> NullableField field
+    NotNull  -> NotNullField field
+
+data FieldDefinition nullability a =
```
aaronallen8455, line 103: I thought about proposing some type synonyms to elide the additional parameter in the common case, possibly even `type FieldDefintion a = FieldDefinitionType NotNull a` so that most existing field definitions won't need changing. I can't find satisfying names though.

aaronallen8455 (approved):

PR #103, opened 2021-04-03
Adds support for composite primary keys

qxjit:
This gives the `tablePrimaryKey` field of `TableDefinition` a rich
`PrimaryKey` data type that can represent composite primary keys and
updates the requisite places to make this work.

The `PrimaryKey` flag is now removed and the migration logic uses the
`tablePrimaryKey` field to determine how to create the primary key.
(Primary key migration is not currently support, and will not be until
we get moved to libpq at which point we may re-evaluate it).

The API remains relatively unchanged in most parts. The streaming
pagination needed to be changed to take a field defintion rather use
`tablePrimaryKey` to determine the ordering of the pagination because
using a composite primary key as the key for pagination will not work
due to ambiguity of ordering correctly on the multiple fields. On the
bright side, this means you can now pagination on any field you choose
rather than being restricted to the primary key field.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
@@ -1,0 +1,84 @@
+{-# LANGUAGE RankNTypes #-}
+module Database.Orville.PostgreSQL.Internal.PrimaryKey
+  ( primaryKeyIn
+  , primaryKeyEquals
+  , primaryKeyDescription
+  , primaryKeyToSql
+  , primaryKey
+  , compositePrimaryKey
+  , primaryKeyPart
+  , mapPrimaryKeyParts
+  ) where
+
+import qualified Data.List as List
+import qualified Database.HDBC as HDBC
+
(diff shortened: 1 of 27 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
```diff
@@ -0,0 +1,84 @@
...
+  ) where
+
+import qualified Data.List as List
+import qualified Database.HDBC as HDBC
+
+import Database.Orville.PostgreSQL.Internal.FieldDefinition (fieldToSqlValue)
+import Database.Orville.PostgreSQL.Internal.Types (PrimaryKey(..), PrimaryKeyPart(..), FieldDefinition(fieldName), NotNull)
+import Database.Orville.PostgreSQL.Internal.Where (WhereCondition, (.<-), (.==), whereAnd, whereOr)
+
+primaryKeyIn :: PrimaryKey key -> [key] -> WhereCondition
```
telser, line 20: I think needing these special functions are the primary downside of this approach.  Maybe some doc strings explaining these equivalents to the FieldDefinition functions is in order?

telser (approved):
I think this is a win, noted the only real point that jumped out at me as not my favorite.

PR #105, opened 2021-04-16
Make DataRowError More Informative

aaronallen8455:
- Changes the result of `sqlTypeFromSql` to `Either String a` instead of
`Maybe a`. The failure reason is included in the RowDataError message.
- When decoding table rows, the RowDataError message includes the primary
key of the row that failed to decode.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
@@ -37,12 +37,12 @@ fieldFromSql field =
   where
     fromSqlValue sql =
       case fieldFromSqlValue field sql of
-        Just a -> Right a
-        Nothing ->
+        Right a -> Right a
+        Left err ->
           Left $
           RowDataError $
           concat
-            ["Error decoding data from column ", fieldName field, " value"]
+            ["Error decoding data from column '", fieldName field, "': ", err]
 
 class ColumnSpecifier col where
   selectForm :: col -> SelectForm
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
@@ -26,7 +26,7 @@ import Database.Orville.PostgreSQL.Internal.Types
 
 convertFromSql :: Convertible SqlValue a => SqlValue -> Either FromSqlError a
 convertFromSql =
-  either (Left . RowDataError . prettyConvertError) Right . safeConvert
+  either (Left . simpleConversionError . prettyConvertError) Right . safeConvert
 
 col :: (ColumnSpecifier col, Convertible SqlValue a) => col -> FromSql a
 col spec = joinFromSqlError (convertFromSql <$> getColumn (selectForm spec))
@@ -40,9 +40,12 @@ fieldFromSql field =
         Right a -> Right a
         Left err ->
           Left $
-          RowDataError $
-          concat
-            ["Error decoding data from column '", fieldName field, "': ", err]
+          RowDataError
+            RowDataErrorDetails
+              { rowErrorReason = err
+              , rowErrorColumnName = fieldName field
+              , rowErrorPrimaryKeys = []
+              }
 
 class ColumnSpecifier col where
   selectForm :: col -> SelectForm
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
@@ -15,6 +15,7 @@ module Database.Orville.PostgreSQL.Internal.SqlType
   , foreignRefType
   , convertSqlType
   , maybeConvertSqlType
+  , eitherConvertSqlType
   ) where
 
 import Control.Monad ((<=<))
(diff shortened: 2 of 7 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
```diff
@@ -37,12 +37,12 @@ fieldFromSql field =
...
       case fieldFromSqlValue field sql of
-        Just a -> Right a
-        Nothing ->
+        Right a -> Right a
+        Left err ->
           Left $
           RowDataError $
           concat
-            ["Error decoding data from column ", fieldName field, " value"]
+            ["Error decoding data from column '", fieldName field, "': ", err]
```
liminalisht, line 45: I played with this change a couple years ago and brought it up to @qxjit . If I remember correctly, the concern was that `err` might contain something sensitive that shouldn't be exposed. Will `err` ever mention the actual value in question that couldn't be decoded @aaronallen8455 ? There may be other reasons that I'm not aware of that this concern is no longer valid.

orville-postgresql/test/ErrorsTest.hs
```diff
@@ -42,6 +42,17 @@ test_errors =
             Right rows ->
               assertFailure
                 ("Expected RowDataError, but got these rows: " ++ show rows)
+
+      , testCase "Includes primary key in error message" $ do
+          run (TestDB.reset schema)
+          void $ run (O.insertRecord virusTable bpsVirus)
+          result <- run $ try (O.selectAll badVirusTable mempty)
+          let expectedErr =
+                "Error decoding data from column 'name': Expected int32, got bytestring (id: SqlInteger 1)"
```
liminalisht, line 51: i like this a lot

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -281,75 +294,107 @@ convertSqlType bToA aToB = maybeConvertSqlType bToA (Just . aToB)
...
+toBoundedInteger :: (Bounded num, Integral num) => Integer -> Either String num
 toBoundedInteger source =
   let truncated = fromInteger source
       upper = toInteger (maxBound `asTypeOf` truncated)
       lower = toInteger (minBound `asTypeOf` truncated)
    in if lower <= source && source <= upper
-        then Just truncated
-        else Nothing
+        then Right truncated
+        else Left "Integral out of range"
```
liminalisht, line 368: this is really nice too

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
qxjit: Yes -- I've always struggled with the concern that adding more information to the row decoding failure will result in accidentally leaking database values that someone considered sensitive at some point. Maybe we can find a way to lock down the error messages to prevent this?

liminalisht (commented):
I really like all the additional information, type mismatches, integral out of bounds, etc., and endorse the use of an `Either String` instead of `Maybe`, with the caveat that I'm concerned about exposing an error encountered from decoding data from a given column *if* that error is going to mention the actual data that it attempted to decode (but could not decode).

qxjit:
@liminalisht @aaronallen8455 What if instead of deriving `Show` for `FromSqlError` we implement it to give a "safe" representation of the error to avoid accidental leakage and the leave it up to applications to extract the details and surface them in away that they consider acceptable based on their own sensitivity context?

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
aaronallen8455: Other than the primary key (which I guess could contain sensitive data...) the values are not included in the error messages. The user could write a custom `FromSql` that includes the value if they intentionally choose to do so.

aaronallen8455:
@qxjit I'm definitely in favor of more structured error messages

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/FromSql.hs
qxjit: @aaronallen8455 Part of my concern is that a user might do so without realizing where the error message ends up getting outputted. For instance, if the error is raised as an exception it could easily end up being returned in an error 500 response by from an API to the end-user of an application. The person composing the error message might not realize that is a possibility and put data in there that think will only ever show up in internal logs, for instance.

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -49,11 +53,11 @@ data SqlType a = SqlType
   , sqlTypeToSql :: a -> HDBC.SqlValue
     -- ^ A function for converting Haskell values of this type into values to
     -- be stored in the database.
-  , sqlTypeFromSql :: HDBC.SqlValue -> Maybe a
+  , sqlTypeFromSql :: HDBC.SqlValue -> Either RowDataErrorReason a
     -- ^ A function for converting values of this are stored in the database
-    -- into Haskell values. This function should return 'Nothing' to indicate
-    -- an error if the conversion is impossible. Otherwise it should return
-    -- 'Just' the corresponding 'a' value.
+    -- into Haskell values. This function should return 'Left String' to
```
qxjit, line 58: This should probably say `Left RowDataErrorReason` instead of `Left String`

qxjit (approved):
:heart:

PR #109, opened 2021-04-26
Add beginnings of SqlMarshaller as a start of 'higher-level' interface.

telser:
- Adds a very much work in progress SqlMarshaller and some small tests
  - Still needs to be tied into `SqlType`
  - Tests are a Hedgehog test tree and are added slightly differently.
- Adds a cabal flag, "ci", and enables much more strict compiler checks with it.
- ci flag is enabled in the test-loop script

### DIFF
--- a/orville-postgresql-libpq/test/Test/SqlMarshaller.hs
+++ b/orville-postgresql-libpq/test/Test/SqlMarshaller.hs
@@ -1,0 +1,33 @@
+module Test.SqlMarshaller
+  ( sqlMarshallerTree
+  ) where
+
+import qualified Hedgehog as HH
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+import Test.Tasty.Hedgehog (testProperty)
+import Test.Tasty (TestTree, testGroup)
+
+import Database.Orville.PostgreSQL.Internal.SqlMarshaller (marshallFromSql)
+
+sqlMarshallerTree :: TestTree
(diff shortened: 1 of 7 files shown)


orville-postgresql-libpq/test/Test/SqlMarshaller.hs
```diff
@@ -0,0 +1,33 @@
...
+  testProperty "Can round trip an int through SqlMarshaller" . HH.property $ do
+    someInt <- HH.forAll generateInt
+    (marshallFromSql (pure someInt) []) HH.=== (Right someInt)
+
+canCombineSqlMarshaller :: TestTree
+canCombineSqlMarshaller =
+  testProperty "Can combine SqlMarshallers with <*>" . HH.property $ do
+    firstInt <- HH.forAll generateInt
+    secondInt <- HH.forAll generateInt
+    marshallFromSql ((pure (+ firstInt)) <*> (pure secondInt)) [] HH.=== (Right (firstInt + secondInt))
```
qxjit, line 30: Would this more more clear written as `fmap (+) (pure firstInt) <*> (pure secondInt)`? Just a though -- I'm not sure which way is better.

qxjit (approved):

aaronallen8455:
How does this differ from `RelationalMap`?

qxjit:
@aaronallen8455 This is the replacement for `RelationalMap` in the `libpq` rewrite.

PR #110, opened 2021-04-26
LibPQ: Adds multi-insert and where clauses to `Expr`

qxjit:
This adds a bit more support to the `Expr` module with types and
constructors for building multi-row insert statements and an initial cut
at basic where comparisons.

Along the way I added a new `readRows` function that reads the rows from
the `LibPQ` result into `SqlValue` lists for further consumption by the
caller. I imagine this will get modified eventually based on the needs
of `SqlMashaller`.

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
@@ -25,6 +25,11 @@ module Database.Orville.PostgreSQL.Internal.Expr
   , andExpr
   , parenthesized
   , comparison
+  , columnEquals
+  , columnGreaterThan
+  , columnLessThan
+  , columnGreaterThanOrEqualTo
+  , columnLessThanOrEqualTo
   , RowValuePredicand
   , columnReference
   , comparisonValue
@@ -135,7 +140,7 @@ andExpr :: BooleanExpr -> BooleanExpr -> BooleanExpr
 andExpr left right =
   BooleanExpr $
     booleanExprToSql left
-    <> RawSql.fromString " OR "
+    <> RawSql.fromString " AND "
     <> booleanExprToSql right
 
 parenthesized :: BooleanExpr -> BooleanExpr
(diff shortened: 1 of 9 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
```diff
@@ -70,19 +107,152 @@ rawTableName :: String -> TableName
...
+  BooleanExpr $
+    booleanExprToSql left
+    <> RawSql.fromString " OR "
+    <> booleanExprToSql right
+
+andExpr :: BooleanExpr -> BooleanExpr -> BooleanExpr
+andExpr left right =
+  BooleanExpr $
+    booleanExprToSql left
+    <> RawSql.fromString " OR "
```
qxjit, line 138: :face_with_head_bandage: I rushed to open this and forgot to change this. I'll add tests for `OR` and `AND` and update this branch before merging.

qxjit: This is fixed in the latest commit on the branch

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlValue.hs
```diff
@@ -45,6 +45,7 @@ import qualified Data.Time as Time
 data SqlValue
   = SqlValue BS.ByteString
   | SqlNull
+  deriving (Show, Eq)
```
telser, line 48: Presumably this is for tests, but I think we want to be careful to say that `SqlValue` isn't safe in general to print.

telser (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlValue.hs
qxjit: Yeah, definitely for tests -- even the `Eq` instance. I wouldn't mind it if `SqlValue` lack that as well. I'll add something so we don't forget to review this, but I suspect we won't know what the best answer looks like until we have more built.

PR #116, opened 2021-05-18
Fix FieldDefinition round-trip test for Double

mitchellvitez:
Only generate doubles under 15 digits long, necessary because of a
PostgreSQL limitation

### DIFF
--- a/orville-postgresql-libpq/test/Test/FieldDefinition.hs
+++ b/orville-postgresql-libpq/test/Test/FieldDefinition.hs
@@ -37,16 +37,12 @@ fieldDefinitionTree pool =
           , roundTripGen = Gen.integral (Range.linearFrom 0 minBound maxBound)
           }
 
-    -- PostgreSQL only stores up to 15 digits of precision, but Haskell values
-    -- can have more precision than that. I created a story to address this
-    -- test to avoid the rest of 'FieldDefinition' getting hung up by this.
-    --
-    -- , testProperty "can round trip a doubleField" . HH.property $ do
-    --   runRoundTripTest pool $
-    --     RoundTripTest
-    --       { roundTripFieldDef = FieldDef.doubleField "foo"
-    --       , roundTripGen = storableDoubleGen
-    --       }
+    , testProperty "can round trip a doubleField" . HH.property $ do
+      runRoundTripTest pool $
+        RoundTripTest
+          { roundTripFieldDef = FieldDef.doubleField "foo"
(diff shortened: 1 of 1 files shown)


orville-postgresql-libpq/test/Test/FieldDefinition.hs
```diff
@@ -98,17 +94,15 @@ fieldDefinitionTree pool =
...
 
--- I tried generating doubles limited to 15 digits, but hedgehog simply gave
--- up because this generator never produced anything. I added a story to track
--- this thread and opted to move forward with getting FieldDefinition in.
---
---storableDoubleGen :: HH.Gen Double
---storableDoubleGen =
---  let
---    isStorable :: Double -> Bool
---    isStorable d = floatDigits d <= 15
```
mitchellvitez, line 109: We could have used a filter like
```
isStorable = (<= 15) . length . filter (/= '.') . takeWhile (/= 'e') . show
```
but I thought it was kind of iffy to rely on `show` for `Double` in that way

telser: +1, definitely would prefer to not use `show` like that.

orville-postgresql-libpq/test/Test/FieldDefinition.hs
```diff
@@ -98,17 +94,15 @@ fieldDefinitionTree pool =
...
---    Gen.filter isStorable $ Gen.double (Range.linearFracFrom 0 (-1000) 1000)
+storableDoubleGen :: HH.Gen Double
+storableDoubleGen =
+  let
+    -- Necessary because PostgreSQL only stores up to 15 digits of precision
+    -- With a 3-digit range, this gives us 12 places after the decimal
+    truncateLongDouble :: Double -> Double
+    truncateLongDouble = (/1e12) . (fromIntegral :: Int -> Double) . round . (*1e12)
+  in
+    flip Gen.subterm truncateLongDouble $ Gen.double $ Range.linearFracFrom 0 (-1000) 1000
```
telser, line 105: That first `$` _could_ be a `.`

telser (approved):
This is great.

orville-postgresql-libpq/test/Test/FieldDefinition.hs
mitchellvitez:
```suggestion
    flip Gen.subterm truncateLongDouble . Gen.double $ Range.linearFracFrom 0 (-1000) 1000
```

PR #118, opened 2021-05-22
LibPQ: Checks for NUL bytes before sending value to LibPQ

qxjit:
This adds a new type, `PGTextFormatValu,e` to track whether we know that
the bytes it contains are all non-NUL (e.g. because we constructed
directly). For any values we do _not_ know are non-NUL, we check them at
the time of building the param list to pass to lib and abort running
the query. For now `executeRow` return `Nothing`, the same as if LibPQ
returned an error. We have another story to start doing better error
handling in general, so I have left actually raising the error about
NUL bytes for that story.

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PGTextFormatValue.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PGTextFormatValue.hs
@@ -1,0 +1,96 @@
+module Database.Orville.PostgreSQL.Internal.PGTextFormatValue
+  ( PGTextFormatValue
+  , NULByteFoundError(NULByteFoundError)
+  , unsafeFromByteString
+  , fromByteString
+  , toByteString
+  , toBytesForLibPQ
+  ) where
+
+import qualified Data.ByteString as BS
+
+{-|
+  A 'PGTextFormatValue' represents raw bytes that will be passed to postgresql
+  via libpq. These bytes must conform to the TEXT format of values that
+  postgresql expects. In all cases postgresql will be allowed to infer the type
+  of the value based on its usage in the query.
+
+  Note that postgresql does not allow NUL bytes in text values, and the LibPQ C
+  library expects text values to be given as NULL-terminated C Strings, so
+  '\NUL' bytes cannot be included in a 'PGTextFormatValue'. If 'fromByteString'
+  is used to construct the 'PGTextFormatValue' (normally what you should do),
+  an error will be raised before libpq is called to execute the query. If
+  'unsafeFromByteString' is used, the caller is expected to ensure that no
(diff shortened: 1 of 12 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PGTextFormatValue.hs
```diff
@@ -0,0 +1,96 @@
...
+
+  Note that postgresql does not allow NUL bytes in text values, and the LibPQ C
+  library expects text values to be given as NULL-terminated C Strings, so
+  '\NUL' bytes cannot be included in a 'PGTextFormatValue'. If 'fromByteString'
+  is used to construct the 'PGTextFormatValue' (normally what you should do),
+  an error will be raised before libpq is called to execute the query. If
+  'unsafeFromByteString' is used, the caller is expected to ensure that no
+  '\NUL' bytes are present. If a '\NUL' byte is included with
+  'unsafeFromByteString', the value passed to the database will be truncated at
+  the '\NUL' byte because it will be interprested as the end of the C String by
```
ysangkok, line 26: s/interprested/interpreted/

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PGTextFormatValue.hs
```diff
@@ -0,0 +1,96 @@
...
+  libpq.
+-}
+data PGTextFormatValue
+  = NoAssumptionsMade BS.ByteString
+  | AssumedToHaveNoNULValues BS.ByteString
```
ysangkok, line 31: Pretty much any statement in source code can be invalidated by the presence of a bug, so what about something simpler like "MayContainNUL" and "WithoutNULs"? But I don't feel strongly, and I don't know the code well, feel free to ignore. Just my two cents :)

ysangkok (commented):
Dunno if this is too much bikeshedding :P

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PGTextFormatValue.hs
qxjit: The reason I didn't chose something `WithoutNULs` is that it could easily imply to a reader that the value is _guaranteed_ to have no nulls because it was checked at runtime, which is not the case. As far as the `PGTextFormatValue` module is concerned it is purely an assumption that there are no nulls. Given that the constructors are not exported from the module anyhow, I erred on the side of precision of meaning, even though it may be a bit wordy.

Basically, if someone ends up needing to muck around in the `PGTextFormatValue` implementation, I want to surface that fact that there is no enforcement regarding nulls in those values to them as soon as possible.

PR #121, opened 2021-05-26
Adds `planEither` to support branching plans

qxjit:
This adds a branching primitive to the `Plan` GADT to allow for
different plans to be given statically for handling the `Left` and
`Right` sides of an either.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
@@ -25,6 +25,7 @@ module Database.Orville.PostgreSQL.Plan
   , apply
   , planMany
   , focusParam
+  , planEither
 
   -- * Bridges from other types into Plan
   , Op.AssertionFailed
@@ -33,6 +34,7 @@ module Database.Orville.PostgreSQL.Plan
   , planOperation
   ) where
 
+import           Data.Either (partitionEithers)
 import qualified Control.Monad.Catch as Catch
 
 import qualified Database.Orville.PostgreSQL.Core as Core
@@ -78,6 +80,10 @@ data Plan scope param result where
 
   PlanMany :: (forall manyScope. Plan manyScope param result)
            -> Plan scope [param] (Many param result)
+
+  PlanEither :: (forall leftScope. Plan leftScope leftParam leftResult)
+             -> (forall rightScope. Plan rightScope rightParam rightResult)
+             -> Plan scope (Either leftParam rightParam) (Either leftResult rightResult)
 
   Bind :: Plan scope param a
        -> (Planned scope a -> Plan scope param result)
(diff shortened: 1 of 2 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
```diff
@@ -327,6 +333,21 @@ focusParam :: (a -> b)
 focusParam focuser plan =
   chain (focuser <$> askParam) plan
 
+
+{-|
+  'planEither' lets you construct a plan branches by executing a different
+  plan for the 'Left' and 'Right' sides of an 'Either' value. When used
+  with a single input paramater only one of the two plans will be used, based on the
+  input parameter. When used on multiple input parameters, each plan will
+  executed only once with all the 'Left' and 'Right' values provided as
```
ysangkok, line 342: "will *be* executed", missing "be"

orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
```diff
@@ -327,6 +333,21 @@ focusParam :: (a -> b)
 focusParam focuser plan =
   chain (focuser <$> askParam) plan
 
+
+{-|
+  'planEither' lets you construct a plan branches by executing a different
```
ysangkok, line 338: "a plan *that* branches", or alternatively "construct branching plans"

orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
ysangkok: If there are multiple input parameters (and parameter is "row"), how can it guaranteed that "each plan" will be executed "once"? Intuitively, I'd think that it chooses the left or right plan independently, for each result? Where does the "once" come into the picture? The only thing that is guaranteed to happen once, is the selection of a plan, for each input, right? If there are zero inputs, surely it can't be choosing at all?

ysangkok: Is it necessary to overload the word "parameter"? In the type signature, it is used for the two types in the Either. But in the comment, it sounds like it is referring to the rows being streamed through the plan.

ysangkok (commented):
Trouble understanding the `planEither` comment.

orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
qxjit: @ysangkok When the plan executes on many parameters at once it uses `partitionEithers` to collect two lists -- all the things from the left and right side of the eithers respectively. Then the left plan is run on all the lefts and the right plan is run on all the rights. Each plan is run just one, providing all the appropriate inputs.

ysangkok: Oooohh I fundamentally misunderstood it. But would it be valid to say "each of the two plans" instead of "each plan"?

qxjit: Yeah, that's a great point. I think "each of the two plans" is much clearer.

qxjit: Regarding your "parameter" feedback, I think we need to address that for `Plan` overall. Currently the value the plan uses as input is referred to as `param` throughout the `Plan` module, including type signatures visible to the user. Maybe calling it `input` throughout the module would be better?

ysangkok: Yes, I think "input" is clearer if it is truly always an input. I think that it is should be possible to not use the same names for value-level and other type-level bindings. So if we have "param" as the common type variable name, it sounds better to use "input" for the value-level value passed in to the plan, to avoid a naming collision. But I recognize that this is separate from this PR.

qxjit: That's interesting. I usually try to match the variable name at the value level with the name at the type level (at least when that makes sense) to avoid confusion. I was imagining possible changing both the `param` type variable and the names at the value level.

qxjit: If any step in a plan encounters an error at any point, the whole plan stops and raises the error, so you will never see any partial plan results.

There's also no lazy IO happening anywhere in the plans (or Orville in general), so there's no streaming possible that could get screwed up by those errors.

telser (approved):

PR #122, opened 2021-05-27
Support years between 0 and 1000

mitchellvitez:
Expands a round-trip test's domain by formatting years to always have four digits


qxjit (approved):
:heart:

PR #123, opened 2021-05-27
LibPQ: Adds OrderBy to Expr

telser:
Adds `OrderByClause` to Expr. This exposes again how easy it is to add new bits of SQL in our current modeling.
But `OrderBy` is, in pure SQL, allowed to have a very wide range of things in it.
To that end this level of abstraction is quite unsafe and potentially invalid SQL is possible.
Creating documentation for the Expr module in general deserves some thought.

The tests have some shared helper functions with the `WhereClause` copied in anticipation of the `Expr` module being broken up and the pieces tested individually.

Left for future work is the handling of NULL first or last, though I suspect that is a relatively straightforward addition.

Also touches up a few minor things:
- Bases off a dockerfile with the latest and greatest stack and debiann
- Adds a stack.yaml file which is a link to the LTS-17.3 specific one
  This makes some tooling which looks for a stack.yaml file a bit easier to work with
- Updates the docker-compose.yml to use the stack.yaml file so we can swap the link as desired without changing the compose file.
- Updates the gitignore to ignore TAGS at the top level instead of only in a directory


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
```diff
@@ -177,6 +185,29 @@ comparison left op right =
...
+  OrderByExpr RawSql
+
+orderByExprToSql :: OrderByExpr -> RawSql
+orderByExprToSql (OrderByExpr sql) = sql
+
+ascendingExpr :: RawSql -> OrderByExpr
+ascendingExpr sql = OrderByExpr $ sql <> RawSql.fromString " ASC"
+
+descendingExpr :: RawSql -> OrderByExpr
+descendingExpr sql = OrderByExpr $ sql <> RawSql.fromString " DESC"
```
qxjit, line 207: Do you think we should have a `newtype OrderByDirection` with `ascending` and `descending` values provided, and then just one `orderByExpr :: RawSql -> OrderByDirection -> OrderByExpr`?

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
```diff
@@ -177,6 +185,29 @@ comparison left op right =
...
+
+addOrderBy :: OrderByExpr -> OrderByExpr -> OrderByExpr
+addOrderBy (OrderByExpr a) (OrderByExpr b) = OrderByExpr (a <> RawSql.fromString ", " <> b)
```
qxjit, line 210: I'm not in live the `addOrderBy` for the name of this -- but my suggestions aren't much better.

The only other suggestion I have is maybe directly represent the sort spec list itself, which would effectively side step the problem by having

`sortSpecList :: [SortSpec] -> SortSpecList`

orville-postgresql-libpq/test/Test/Expr.hs
```diff
@@ -25,94 +25,184 @@ data FooBar =
...
+  describe "OrderBy" $ do
+    it "ascendingExpr sorts a text column" $ do
+      runOrderByTest pool $
+        OrderByTest
+          { orderByValuesToInsert = [FooBar 1 "dog", FooBar 2 "dingo", FooBar 3 "dog"]
+          , orderByExpectedQueryResults = [FooBar 2 "dingo", FooBar 1 "dog", FooBar 3 "dog"]
+          , orderByClause =
+              Just . Expr.orderByClause $
+                Expr.ascendingExpr
+                  (Expr.columnNameToSql barColumn)
```
qxjit, line 127: I would be nice to have a solution for the user needing to call `columnNameToSql` for what is the surely to be the most common use case -- sorting by a column.

qxjit (approved):
I added a few thoughts. I'm sure this will evolve over time as we continue to explore the design of the whole `Expr` module, so take or leave them on this PR as you wish.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr.hs
telser: I toyed some with having a type that carried the direction with it but wasn't happy with where that was going either so I punted back to this. Happy to revisit that though.

orville-postgresql-libpq/test/Test/Expr.hs
telser: Agreed. I think we'll probably want some number of convenience functions for that kind of thing built around these more raw level functions that exist now.

PR #126, opened 2021-06-01
LibPQ: Surfaces SQL Execution Errors

qxjit:
This raises errors that can happen during exection as `Exception` values
in the `IO` Monad. These can be errors detected on the client side (e.g.
connection closed, null bytes found in input parameters) or returned
from the database sql (e.g. sql syntax error).

Once we start adding the high-level interfaces (in particular, using
orville within monad stacks) we may decide we want this lower level
interface to offer non-IO based error handling. We can address that
question once we get to it.

Side note: this removes the `stack.yaml` symlink because `docker-compose
up` was failing with a complaint that it could not find the file.

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
@@ -7,21 +7,28 @@ License   : MIT
 module Database.Orville.PostgreSQL.Connection
   ( Connection
   , Pool
+  , ConnectionUsedAfterCloseError
+  , ConnectionError
+  , SqlExecutionError(..)
   , createConnectionPool
   , executeRaw
   , executeRawVoid
   ) where
 
-import Control.Concurrent (threadWaitRead, threadWaitWrite)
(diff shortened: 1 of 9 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
```diff
@@ -165,13 +238,44 @@ mkInferredTextParam mbValue =
...
+
+          Just libPQMsg ->
+            case Enc.decodeUtf8' libPQMsg of
+              Right decoded ->
+                T.unpack decoded
+
+              Left decodingErr ->
+                "Error decoding libPQ messages as utf8: " <> show decodingErr
+     in
+      connectionErrorMessage err <> libPQErrorMsg
```
aaronallen8455, line 263: Add a delimiter here?

aaronallen8455 (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Connection.hs
qxjit: Thank you -- I missed that I had accidentally removed then in translating from the older code.

PR #127, opened 2021-06-05
Breaks apart Expr module

telser:
This tries to find a balance between logical groupings of code and premature abstraction.
As with everything else I highly suspect we'll revist this many times as knowledge and needs shift.
The heart of this is to make the code a bit easier to digest and thus modify.

Some more details of changes:
- Makes Expr a re-exporter module
- Adds an Expr/Query hierarchy with the relevant bits for creating query expressions
- Adds an Expr/InsertExpr module with the insertion expression bits
- Adds an Expr/OrderBy hierarchy with all the OrderBy express code
- Adds an Expr/Where heirarchy with all the Where expression code
- Adds an Expr/Name heirarchy containing the Column and Table naming code
- Adds an Expr/FieldDefinition module
- Updates the years in modules to say 2021

FieldDefinition currently _feels_ a bit out of place with the others.
But I suspect future work with tables will likely resolve this.


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/FieldDefinition.hs
```diff
@@ -0,0 +1,114 @@
...
+  , bigSerial
+  , bigInt
+  , serial
+  , int
+  ) where
+
+import Database.Orville.PostgreSQL.Internal.Expr.Name
+  ( ColumnName
+  , columnNameToSql
+  )
```
qxjit, line 29: I really dislike multi-line `import` style. Can we agree on something different?

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/FieldDefinition.hs
```diff
@@ -0,0 +1,114 @@
...
+import Database.Orville.PostgreSQL.Internal.RawSql
+  ( RawSql
+  , fromString
+  )
+
+newtype FieldDefinition =
+  FieldDefinition RawSql
```
qxjit, line 36: We should talk about how we want to name this. I don't know if it would be more clear to call it `ColumnDefinition` to make it distinct from the `FieldDefinition` type that is user with `SqlMarshaller` or if that would be more confusing since they're analogous things operating at two different tiers of the API.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/FieldDefinition.hs
```diff
@@ -0,0 +1,114 @@
...
+  , int
+  ) where
+
+import Database.Orville.PostgreSQL.Internal.Expr.Name
+  ( ColumnName
+  , columnNameToSql
+  )
+import Database.Orville.PostgreSQL.Internal.RawSql
+  ( RawSql
+  , fromString
```
qxjit, line 32: I'd rather see the function from `RawSql` get used qualified here, so that the code says `RawSql.fromString`.

qxjit (changes requested):
I think this organization is great. We just need to come to some stylistic agreements about import formatting and when to qualify imports.

qxjit (approved):

Issue #124, opened 2021-05-28
findRecordsBy, findRecordsByCached, and groupBy' all reverse the order of the lists they return


liminalisht:
On the `0.9_development` branch (and apparently also on `master`), `findRecordsBy` and `findRecordsByCached` do not preserve the ordering specified in any order by clauses. Instead they appear to reverse that ordering, due to a bug in `groupBy'`.

Let's say I write:
```
findRecordsBy someTable someField
$ where_
    (whereAnd
      [ ...
      ]
    )
<> O.order field1 O.Descending
<> O.order field2 O.Descending
```

I would expect this to return an `m (Map key [record])`, where the `[record]` in the `Map` is sorted in the way I defined in the order clauses, namely field1 desc, field2 desc.

What happens instead is that the order of this list is reversed.

This is because 
```
findRecordsBy ::
     (Ord fieldValue, MonadOrville conn m)
  => TableDefinition readEntity writeEntity key
  -> FieldDefinition nullability fieldValue
  -> SelectOptions
  -> m (Map.Map fieldValue [readEntity])
findRecordsBy tableDef field opts = do
  let builder = (,) <$> fieldFromSql field <*> tableFromSql tableDef
      query = selectQuery builder (fromClauseTable tableDef) opts
  Map.groupBy' id <$> runSelect query
```
and 
```
findRecordsByCached ::
     (Ord fieldValue, MonadThrow m, MonadOrville conn m)
  => TableDefinition readEntity writeEntity key
  -> FieldDefinition nullability fieldValue
  -> SelectOptions
  -> QueryCached m (Map.Map fieldValue [readEntity])
findRecordsByCached tableDef field opts = do
  let builder = (,) <$> fieldFromSql field <*> tableFromSql tableDef
  rows <- selectCachedRows tableDef opts
  Map.groupBy' id <$> unsafeLift (decodeSqlRows builder rows)
```
both rely on
```
groupBy' :: Ord k => (a -> (k, v)) -> [a] -> Map.Map k [v]
groupBy' mkEntry as = Map.fromListWith (++) (map mkListEntry as)
  where
    mkListEntry a =
      let (k, v) = mkEntry a
       in (k, [v])
```
which take your `as :: [a]`, and maps each `a` to a key/value pair, where the value is always a singleton list `[v]`. Then `Map.fromListWith (++)` is invoked on that.

Notice that this will end up reversing the elements of the list. If my `as` here are `[2, 3]`, I end up with `[3, 2]`
```
[...]

PR #128, opened 2021-06-08
Fixes ordering being inverted in queries that use map helpers.

telser:
This is to address #124 

Tests added are adapted from the `WhereCondition` test module for expediency.

### DIFF
--- a/orville-postgresql/src/Data/Map/Helpers.hs
+++ b/orville-postgresql/src/Data/Map/Helpers.hs
@@ -16,7 +16,7 @@ groupBy keyFunc = groupBy' mkEntry
     mkEntry a = (keyFunc a, a)
 
 groupBy' :: Ord k => (a -> (k, v)) -> [a] -> Map.Map k [v]
-groupBy' mkEntry as = Map.fromListWith (++) (map mkListEntry as)
+groupBy' mkEntry as = Map.fromListWith (flip (++)) (map mkListEntry as)
   where
     mkListEntry a =
       let (k, v) = mkEntry a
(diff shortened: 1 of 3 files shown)


orville-postgresql/src/Data/Map/Helpers.hs
```diff
@@ -16,7 +16,7 @@ groupBy keyFunc = groupBy' mkEntry
     mkEntry a = (keyFunc a, a)
 
 groupBy' :: Ord k => (a -> (k, v)) -> [a] -> Map.Map k [v]
-groupBy' mkEntry as = Map.fromListWith (++) (map mkListEntry as)
+groupBy' mkEntry as = Map.fromListWith (flip (++)) (map mkListEntry as)
```
aaronallen8455, line 19: Can we use `DList` (or equivalent) instead to avoid left associating `++`?

telser: We could round trip through a `DList` I suppose, and exploring something like `DList` more could be great for the libpq version, but I certainly don't want to introduce a breaking api change with a bug fix like this.

aaronallen8455: Yes, round tripping is what I mean, no api change

qxjit: Agree that we avoid flipping `++` and putting the long list on the left side of it.  Round-tripping through `DList` seems like a reasonable and quick thing to do though.

aaronallen8455 (approved):

qxjit (approved):

PR #130, opened 2021-06-09
monomorphized param in `planSelect`

ysangkok:
Submitting this since it bit @mitchellvitez (https://github.com/exfreight-zeta/schmods/pull/2483/commits/ef43a0239bf12c7c1a01180f6430cab0215a3fb1) and @eightball suggested it shouldn't be polymorphic. ([slack thread](https://flipstonetech.slack.com/archives/C6286FS3D/p1623247932499700?thread_ts=1623247871.499300&cid=C6286FS3D))


qxjit:
I'm game to merge this and see if we like it. Notably it changes the ergonomics of using it because it will force you to use something like `focusParam` to get the `()` that the plan now requires as input, where as the ploymorphic version can be used without needing that extra step.

qxjit (approved):

PR #132, opened 2021-06-09
Brings 0.9-development and 'master' branches together 

telser:
This prefers changes from the 0.9 branch. 

Also included here is a version bump for the HDBC backed code to '0.9.0.0' and the libpq backed code to '0.10.0.0'. This way we can have a stable branch for the HDBC code with a new revision number based from the primary branch. Also allows us to continue the libpq development on the primary branch.


qxjit (approved):

telser:
This didn't keep all of the history, so master was pushed back and we recreated the spirit of this in #134 which _does_ have all the history.

PR #133, opened 2021-06-10
LibPQ: Adds marshallEntityToSql

qxjit:
This adds a `marshallEntityToSql` function that can be used to convert a
Haskell entity to the SQL fields described by the marshaller. This is
implemented in as a fold-style function to allow the caller to build
whatever result type is required for their context.

Along the way I introduced the `FieldName` type to help make the
signature of `marshallEntityToSql` easier to understand. I went ahead
and made it a wrapper around `ByteString` rather than `String` since
that is mostly how we used the value.


qxjit:
Closing this while we sort out master branch.

PR #138, opened 2021-06-14
GHC 9 compatibility

ysangkok:
The code is not building with GHC 9, which is already out and used by some people.
Some people may already potentially be using GHC 9, no need to exclude them.

The fix is of this commit is called "manual eta-expansion" and is
mentioned in the "Simplify Subsumption" proposal.

It still works with GHC 8.6.

https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Execute.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Execute.hs
@@ -12,7 +12,9 @@ import Database.Orville.PostgreSQL.Internal.Monad
 
 executingSql :: MonadOrville conn m => QueryType -> String -> IO a -> m a
 executingSql queryType sql action = do
-  runningQuery <- ormEnvRunningQuery <$> getOrvilleEnv
+  runningQuery <-
+    (\queryType -> ormEnvRunningQuery queryType)
+      <$> getOrvilleEnv
   liftIO $ runningQuery queryType sql (catchSqlErr sql action)
 
 catchSqlErr :: String -> IO a -> IO a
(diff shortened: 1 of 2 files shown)


qxjit:
Is there any warning we can turn on in prior GHC versions that would let us know if we add more of these before we have an official GHC 9 build for orville?

qxjit (approved):

ysangkok:
@qxjit I don't think there is any warning that would detect this. I was reading looking through comments on [the GHC PR implementing "Simplify Subsumption"](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2600) and there is no mention of backporting anything. Simon PJ also mentions that "our ambiguity check is a bit of a heuristic", I am not sure if that is relevant. But if it is relevant here, it would detract from backporting changes since a heuristic is probably not specified and therefore also difficult to backport.

qxjit:
@ysangkok I less concerned about it being backported as I am simply that we have no check in orville right now that we don't break GHC9 again. Often times when GHC breaks things they provide a warning options to turn on in prior versions to check for it ahead of time. 

If we break things it will be easy to fix in the future though.

ysangkok:
@qxjit Right, but given that it is such a massive change to GHC (look at the size of the PR), I was thinking that a warning would need to implement the same code, making it similar to a backport. In our case, it the error looked like it was caused by an implicitly moved `forall`, but that is just in this case, and I don't even know if it would be easy for them to check for a "moved `forall`". So I mention the heuristics since I think that is the root cause: There was no specification of the subsumption previously, and while you could argue that there is now (the proposal), calculating the difference between the two could be difficult.

PR #139, opened 2021-06-14
LibPQ: Adds minimal port of TableDefinition

qxjit:
This adds a minimal port of `TableDefinition`, containing just the table
name, marshaller and primary key. It includes functions to generate SQL
expressions for table creation, record insertion and querying.

`SqlMarshaller` needed some API tuning to get everything that
`TableDefinition` needed for its job. The "marshall to sql" function
became simply a fold over the fields in the `SqlMarshaller`, which is
used in a couple of different ways for building the queries to support
table operations.

We may decide that we don't want to export the `TableDefinition`
constructor (so that we can add record fields without it being a
breaking api change), but I deferred that decision for the moment in
favor of getting this basic functionality working.

### DIFF
--- a/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
+++ b/orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
@@ -6,6 +6,7 @@ module Database.Orville.PostgreSQL.Internal.PrimaryKey
     primaryKeyDescription,
     primaryKeyToSql,
     primaryKey,
+    PrimaryKeyPart,
     compositePrimaryKey,
     primaryKeyPart,
     mapPrimaryKeyParts,
@@ -28,9 +29,9 @@ data PrimaryKey key
   = PrimaryKey (PrimaryKeyPart key) [PrimaryKeyPart key]
 
 {- |
-  An internal helper type representing a single part of a composite primary
-  key. This is not export in part to hide the use of GADTs, and also simply
-  to avoid leaking implementation details.
+  A 'PrimaryKeyPart' describes one field of a composite primary key. Values
+  are built using 'primaryKeyPart' and then used with 'compositePrimaryKey'
+  to build a 'PrimaryKey'
 -}
 data PrimaryKeyPart key
   = forall part. PrimaryKeyPart (key -> part) (FieldDefinition NotNull part)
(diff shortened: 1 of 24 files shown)


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlMarshaller.hs
```diff
@@ -56,33 +56,53 @@ instance Applicative (SqlMarshaller a) where
...
   SqlMarshaller writeEntity readEntity ->
   result ->
-  (FieldName -> SqlValue -> result -> result) ->
-  writeEntity ->
+  ( forall a nullability.
+    FieldDefinition nullability a ->
+    (writeEntity -> a) ->
+    result ->
+    result
+  ) ->
```
aaronallen8455, line 72: I think it'd be useful to have a type synonym for this.
```
type FieldFold writeEntity result = ...
```

qxjit: @aaronallen8455 That's a great suggestion. I'll add that.

aaronallen8455 (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SqlMarshaller.hs
aaronallen8455: The documentation is very nice, btw.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/RawSql.hs
```diff
@@ -16,6 +16,12 @@ module Database.Orville.PostgreSQL.Internal.RawSql
     intercalate,
     execute,
     executeVoid,
+
+    -- * Fragmants provided for convenience
+    space,
```
telser, line 21: These are going to be great, I think, for readability.

orville-postgresql-libpq/test/Test/TableDefinition.hs
```diff
@@ -0,0 +1,112 @@
...
+          dropAndRecreateTestTable pool fooTable
+          RawSql.executeVoid pool (Expr.insertExprToSql insertFoo)
+          RawSql.executeVoid pool (Expr.insertExprToSql insertFoo)
+
+        case result of
+          Right () -> do
+            HH.footnote "Expected 'executeVoid' to return failure, but it did not"
+            HH.failure
+          Left err ->
+            sqlExecutionErrorSqlState err === Just (B8.pack "23505")
```
telser, line 66: Maybe we will want to turn the 23505 into a helper definition at some point?

qxjit: @telser Yeah, we might expose a `SqlState` module with a bunch of known values.

telser (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/PrimaryKey.hs
```diff
@@ -90,11 +91,11 @@ compositePrimaryKey =
   PrimaryKey
 
 {- |
-  'primaryKeyPart' builds on section of a composite primary key based on the
-  field definition that corresponds to that column of the primary key. The
-  function given is used to decompose the Haskell value for the composite key
-  into the individual parts so they can be converted to sql for things like
-  building 'WhereCondition'
+  'primaryKeyPart' builds building block for a composite primary key based a
```
telser, line 94: Nitpick: Should this say "builds a building block for.." or "is a building block for"?

PR #142, opened 2021-06-17
Add planList

mitchellvitez:


orville-postgresql/src/Database/Orville/PostgreSQL/Plan.hs
```diff
@@ -321,6 +322,16 @@ planMany :: (forall manyScope. Plan manyScope param result)
...
 
+{-|
+  'planList' lifts a plan so both its param and result become lists.
+  This saves you from having to fmap in 'Many.elems' when all you want back
+  from a 'Many' is the list of results inside it.
+-}
+planList :: (forall scope. Plan scope param result)
+         -> Plan listScope [param] [result]
+planList plan =
+  Many.elems <$> planMany plan
```
telser, line 333: If we do this in the eta reduced style of `fmap Many.elems . planMany` then I think GHC has a better time in the optimization passes. IIRC this has more impact on functions like this, around specialization.

telser (approved):

qxjit:
@telser @mitchellvitez I think we probably want this on the `Orville-0.9` branch, don't we?

PR #143, opened 2021-06-17
LibPQ: Adds MonadOrville and friends

qxjit:
This adds `MonadOrville`, `MonadOrvilleControl` and `HasOrvilleState` in
their basic forms so that libpq orville can be used from an application
monad. Currently `insertRecord` is provided as an example to test this.


qxjit:
I saw that formolu message from the build locally -- but I didn't realize it was exiting with an error code. It looked like it still formatted the file. I'll look into what is offending it before merging anything.

telser (approved):

PR #144, opened 2021-06-19
LibPQ: Adds full support for field nullability

qxjit:
This adds the rest of the field nullability support from the HDBC
version. The main different is that I chose to not to export the GADT
for nullability outside the module itself. I think this probably makes
for a cleaner and less confusing API overall. The `FieldNullability`
type is the equivalent of `NullabilityCheck` from the HDBC code,
allowing users of `FieldDefinition` to recover the type-level
nullability information without needing to turn on GADTs.


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/FieldDefinition.hs
```diff
@@ -48,6 +49,7 @@ import qualified Database.Orville.PostgreSQL.Internal.RawSql as RawSql
 import Database.Orville.PostgreSQL.Internal.SqlType (SqlType)
 import qualified Database.Orville.PostgreSQL.Internal.SqlType as SqlType
 import Database.Orville.PostgreSQL.Internal.SqlValue (SqlValue)
```
telser, line 51: How do you feel about going ahead and just using the qualified import in cases like this?

telser (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/FieldDefinition.hs
qxjit: I guess I don't mind changing it in this case since there aren't very many references, but I'm still torn about it in cases where the type name is used pervasively through the file.

PR #148, opened 2021-06-21
HDBC: Drop Index only if it exists during migration.

### LINKED ISSUE #141: Drop index if exists
Migrations that include `DropIndex` will fail if any of the columns that the index in question references have already been dropped. This is because when one drops a column in postgresql, postgres will also drop indexes and constraints on your behalf.

Consider amending `DropIndex` to perform a `DROP INDEX IF EXISTS`.

telser:
This fixes #141

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/MigrateIndex.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/MigrateIndex.hs
@@ -39,4 +39,6 @@ createIndexPlan indexDef schemaState = do
 dropIndexPlan :: String -> SchemaState -> Maybe MigrationPlan
 dropIndexPlan name schemaState = do
   guard (schemaStateIndexExists name schemaState)
-  pure $ migrationDDLForItem (DropIndex name) ("DROP INDEX " ++ name)
+  -- Here we drop the index only if it exists so that during the migration plan
+  -- should something else prior cause the index to be dropped we will not cause an error.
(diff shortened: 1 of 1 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/MigrateIndex.hs
```diff
@@ -39,4 +39,6 @@ createIndexPlan indexDef schemaState = do
 dropIndexPlan :: String -> SchemaState -> Maybe MigrationPlan
 dropIndexPlan name schemaState = do
   guard (schemaStateIndexExists name schemaState)
-  pure $ migrationDDLForItem (DropIndex name) ("DROP INDEX " ++ name)
+  -- Here we drop the index only if it exists so that during the migration plan
+  -- should something else prior cause the index to be dropped we will not cause an error.
+  pure $ migrationDDLForItem (DropIndex name) ("DROP INDEX IF EXISTS" ++ name)
```
aaronallen8455, line 44: don't you need a space after EXISTS here?

aaronallen8455 (approved):

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/MigrateIndex.hs
telser: mmm I must have deleted it after running my local test, somehow. It's fixed now.

PR #151, opened 2021-06-23
LibPQ: Adds SelectOptions with WhereCondition

qxjit:
This begins porting the `SelectOptions` type, only supporting
`WhereCondition`s for the moment.

I opted to forego the old version `WhereCondition` that modeled where
conditions explicity in Haskell in favor of just a newtype wrapper
around `BooleanExpr`. This made it much easier to implement and provided
an obvious and natural interface for incorporated expression that
Orville does not support directly into the higher-level API.

I also introduced a `ToRawSql` class that I think can help us make some
function more re-useable. I only added it in a few places for the
moment, but if we like this direction then we can adjust the rest of the
`Expr` interface to orient around it.

### DIFF
--- a/orville-postgresql-libpq/orville-postgresql-libpq.cabal
+++ b/orville-postgresql-libpq/orville-postgresql-libpq.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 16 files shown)


qxjit:
Closing this and will re-open once I figure out what happened wit formatting. It's no longer in my build.sh script!

qxjit:
Re opened as https://github.com/flipstone/orville/pull/152

PR #152, opened 2021-06-23
LibPQ: Adds SelectOptions with WhereCondition

qxjit:
This begins porting the `SelectOptions` type, only supporting
`WhereCondition`s for the moment.

I opted to forego the old version `WhereCondition` that modeled where
conditions explicity in Haskell in favor of just a newtype wrapper
around `BooleanExpr`. This made it much easier to implement and provided
an obvious and natural interface for incorporated expression that
Orville does not support directly into the higher-level API.

I also introduced a `ToRawSql` class that I think can help us make some
function more re-useable. I only added it in a few places for the
moment, but if we like this direction then we can adjust the rest of the
`Expr` interface to orient around it

I also included a commit to re-add formatting to build.sh. Somehow I accidentally removed it while retabbing the file, ironically.


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SelectOptions/WhereCondition.hs
```diff
@@ -0,0 +1,135 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Database.Orville.PostgreSQL.Internal.SelectOptions.WhereCondition
+  ( WhereCondition,
+    whereEquals,
+    whereNotEquals,
+    whereGreaterThan,
+    whereLessThan,
+    whereGreaterThanOrEqualTo,
+    whereLessThanOrEqualTo,
```
aaronallen8455, line 10: I'm in favor of have operator aliases for these.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SelectOptions/WhereCondition.hs
```diff
@@ -0,0 +1,135 @@
...
+whereAnd :: NonEmpty WhereCondition -> WhereCondition
+whereAnd =
+  foldParenthenizedExprs Expr.andExpr
+
+{- |
+  Combines multiple 'WhereCondition's together using 'OR.
+-}
+whereOr :: NonEmpty WhereCondition -> WhereCondition
+whereOr =
+  foldParenthenizedExprs Expr.orExpr
```
aaronallen8455, line 120: I think these should mirror the familiar `and` and `or` functions and not be restricted to NonEmpty lists.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SelectOptions/WhereCondition.hs
qxjit: We will add some eventually, but we're building the entire 0.10 interface as "function names first" and then will decide how/where to add operators once we feel like the API is pretty solid and ergonomic without them.

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/SelectOptions/WhereCondition.hs
qxjit: Personally, I think the empty-list cases even more confusing in the case of building SQL expressions than they are when dealing with simple boolean values. Using `NonEmpty` here allows us to remove that confusion entirely.

In particular, if you consider `whereOr []`, it's ambiguous whether the empty list of where conditions should mean "I don't want any results returned" or "I have no conditions to search by, so please return everything". I can actually think of more use-cases for the latter off hand (e.g. building search expressions from user input) than the former, but that is the opposite of what the familiar `or` would dictate.

Even in the case of `whereAnd`, it can be non-obvious what you wish would happen in terms of the SQL itself -- did you want no where condition to be put on the query at all, or a `WHERE TRUE` to be appended? 

If we discover this is horribly un-ergonomic in practice I'm definitely open to revisiting.

aaronallen8455: I don't agree about it being ambiguous. `whereOr` should mean "at least one of these conditions must hold" and this is clearly going to be false for an empty list of conditions. I can see how this might be surprising to some but I don't think it should be guarded against at the expense of ergonomics.

qxjit: I see your point, though I don't think it's quite so clear cut. I've created a story internally for us to take another look at this to make sure we get the right API in the end. I'd like to defer that decision to a later point for now so we can get more of the API in place over LibPQ.

PR #153, opened 2021-06-29
LibPQ: Swaps tests to be hedgehog only and adds ghc9

telser:
- Swaps test suite to be completely Hedgehog based
  - This reduces the test suite dependencies from 82 to 58 (according to `stack ls dependencies`)!
- Updates code to build with ghc 9.0.1
  - Bumps upper bound on `dlist`
  - Changes a type signature on a local function (which now requires `RankNTypes` to be turned on for the module
- Adds a nightly stackage resolver to test with ghc9.0.1 with
- Adds ghc9 to circleci
- Corrects list of `tested-with` in package.yaml/cabal
- Updates test suite to use more qualified imports


orville-postgresql-libpq/test/Main.hs
```diff
@@ -3,40 +3,38 @@ module Main
...
+import qualified Test.Connection as TestConnection
+import qualified Test.Expr.OrderBy as ExprOrderBy
+import qualified Test.Expr.Where as ExprWhere
+import qualified Test.FieldDefinition as FieldDefinition
+import qualified Test.RawSql as RawSql
+import qualified Test.RecordOperations as RecordOperations
+import qualified Test.SelectOptions as SelectOptions
+import qualified Test.SqlMarshaller as SqlMarshaller
+import qualified Test.SqlType as SqlType
+import qualified Test.TableDefinition as TableDefinition
```
qxjit, line 20: :heart: I've considered qualifying these test imports several times. This was the perfect PR to do it in.

orville-postgresql-libpq/test/Test/Connection.hs
```diff
@@ -1,125 +1,135 @@
...
+                    connection
+                    (B8.pack "SELECT $1::text = $2::text")
+                    [ Just $ PGTextFormatValue.fromByteString notNulBytes
+                    , Just $ PGTextFormatValue.unsafeFromByteString notNulBytes
+                    ]
+
+                LibPQ.getvalue' result 0 0
+
+            value HH.=== Just (B8.pack "t")
+        )
```
qxjit, line 50: I could definitely see us creating a helper that took a `String` build this tuple for us

orville-postgresql-libpq/test/Main.hs
```diff
@@ -3,40 +3,38 @@ module Main
...
+    sequence
+      [ TestConnection.connectionProperties pool
+      , RawSql.rawSqlProperties
+      , TableDefinition.tableDefinitionProperties pool
+      , SqlMarshaller.sqlMarshallerProperties
+      , FieldDefinition.fieldDefinitionProperties pool
+      , RecordOperations.recordOperationsProperties pool
+      , ExprWhere.whereSpecs pool
+      , ExprOrderBy.orderBySpecs pool
+      , SqlType.sqlTypeSpecs pool
```
qxjit, line 37: Should we rename the ones that have `Specs` to `Properties`, or is there a convention that you're using for when one is `Specs` vs `Properties`?

orville-postgresql-libpq/test/Test/Property.hs
```diff
@@ -0,0 +1,16 @@
...
+import qualified GHC.Stack as CallStack
+import qualified Hedgehog as HH
+
+singletonProperty :: CallStack.HasCallStack => HH.PropertyT IO () -> HH.Property
+singletonProperty = HH.withTests 1 . HH.property
+
+-- data Groups =
+--   Groups { runSequentially :: [HH.Group]
+--          , runParallel :: [HH.Group]
+--          }
```
qxjit, line 16: Oops, some dead code in here

qxjit (approved):

PR #154, opened 2021-07-01
Add GroupBy

mitchellvitez:
- Adds `GroupBy` (wrapping raw sql) to `Expr`
- Adds skeleton tests which can be expanded as `GroupByExpr` expands


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/GroupBy/GroupByExpr.hs
```diff
@@ -0,0 +1,27 @@
...
+
+groupByExprToSql :: GroupByExpr -> RawSql.RawSql
+groupByExprToSql (GroupByExpr sql) = sql
+
+appendGroupBy :: GroupByExpr -> GroupByExpr -> GroupByExpr
+appendGroupBy (GroupByExpr a) (GroupByExpr b) =
+  GroupByExpr (a <> RawSql.comma <> b)
+
+groupByExpr :: RawSql.RawSql -> GroupByExpr
+groupByExpr sql =
```
qxjit, line 26: I added a note to the story to add `OrderBy` to `SelectOptions` that we may want a slighty nicer interface than one that simply takes `RawSql`. My guess is that we'll at least want something like `groupByColumn :: ColumnName -> GroupByExpr` to make it clear how to group by a column specifically, since that's by far the most common case.

orville-postgresql-libpq/test/Test/Expr.hs
```diff
@@ -189,11 +190,69 @@ runOrderByTest pool test =
...
-          (Expr.tableExpr exprTestTable Nothing (orderByClause test))
+          (Expr.tableExpr exprTestTable Nothing (orderByClause test) Nothing)
 
     rows <- ExecResult.readRows result
     rows `shouldBe` mkOrderByTestExpectedRows test
 
+groupBySpecs :: Pool Connection -> Spec
+groupBySpecs pool =
+  describe "GroupBy" $ do
+    it "appendGroupBy causes grouping on both clauses" $ do
```
qxjit, line 201: Oops, you're going to need to merge `master` in and update this test to use HedgeHog rather than HSpec

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/GroupBy/GroupByExpr.hs
```diff
@@ -0,0 +1,27 @@
...
+
+import qualified Database.Orville.PostgreSQL.Internal.RawSql as RawSql
+
+newtype GroupByExpr = GroupByExpr RawSql.RawSql
+
+groupByExprToSql :: GroupByExpr -> RawSql.RawSql
+groupByExprToSql (GroupByExpr sql) = sql
+
+appendGroupBy :: GroupByExpr -> GroupByExpr -> GroupByExpr
+appendGroupBy (GroupByExpr a) (GroupByExpr b) =
```
qxjit, line 22: @mitchellvitez @telser Any thoughts about providing `Semigroup` instances for a number of these expressions that have natural combination semantic? I'd still want to have `appendGroupBy`, but allowing `<>` to be used if the user chooses to would be nice.

mitchellvitez: I'm in favor, allowing reuse of known concepts is nice and I don't think Semigroup introduces much overhead

aaronallen8455:
Is there a plan for how to handle aggregation of columns not in the group by clause? Postgres will reject queries that have selected columns not in the group by clause if they are not being aggregated somehow.

qxjit:
@aaronallen8455 the group by support in Orville has historically been used in conjunction with the lower-level Orville functions that let you write your own select statement.  That said, I would say that for the `0.10` release on LibPQ we _do_ plan to add some support for aggregate function helpers.

qxjit (approved):

qxjit:
@mitchellvitez Let's go ahead and get this merged

PR #155, opened 2021-07-02
LibPQ: Adds entity operations for find and update

qxjit:
This adds `updateEntity` and `findEntity` and a couple of friends for
basic entity operations, along with the `Expr` additions required to
support them.


orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/Query/SelectList.hs
```diff
@@ -25,7 +25,7 @@ selectColumns columnNames =
   SelectList $
     RawSql.intercalate
       RawSql.comma
-      (fmap columnNameToSql columnNames)
+      (map RawSql.toRawSql columnNames)
```
telser, line 28: Should this be another place that is `NonEmpty`? Either way, if we stick to `fmap` it certainly makes performing those changes easier.

qxjit: I don't think `RawSql.intercalate` should take a `NonEmpty` list, if that's what you're asking.

As far as `fmap` vs `map` I guess I don't care that much. Obviously switching between them is not a big deal either way. I just like keeping things simple. I can change it though.

telser: I meant `selectColumns`. Definitely agree that `RawSql.intercalate` being defined on `List` makes more sense.

telser (approved):

orville-postgresql-libpq/src/Database/Orville/PostgreSQL/Internal/Expr/Query/SelectList.hs
qxjit: Ah, I see. Maybe we should consider changing `selectColumns` to take `NonEmpty`, but that's definitely inconvenient at the moment, because `SqlMarshaller`  doesn't guarantee that it has any fields at all (e.g. `pure ()` is a perfectly fine `SqlMarshaller` today).

PR #156, opened 2021-07-07
Rename Database.Orville to Orville in libpq

mitchellvitez:
Reduces import lengths and namespaces Orville differently

This change is just:

```sh
cd orville-postgresql-libpq
find . -type f | xargs sed -i '' -e 's/Database.Orville/Orville/g'
mv src/Database/Orville src
rm -r src/Database
```


qxjit (approved):
:heart:

PR #160, opened 2021-07-26
Fix mkTableDefiniton to mkTableDefinition

OwenGraves:


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/TableDefinition.hs
```diff
@@ -47,7 +47,7 @@ data TableDefinition key writeEntity readEntity = TableDefinition
   Constructs a new 'TableDefinition' with the minimal fields required for
   operation.
 -}
-mkTableDefiniton ::
+mkTableDefinition ::
```
qxjit, line 50: @OwenGraves Looks like you missed renaming the implementation below this type signature

qxjit (changes requested):

qxjit (approved):

telser (approved):

PR #162, opened 2021-07-29
Make SqlValue more opaque by removing Show, Eq

mitchellvitez:
Remove these instances on `SqlValue`, but keep tests running by passing`SqlValue`s through `toText` first. If needed in the future, we could further process those `Text`s

It's pretty likely there's a better place for the `sqlValuesToText` helper to live, happy to move it.

### DIFF
--- a/orville-postgresql-libpq/test/Test/Expr/OrderBy.hs
+++ b/orville-postgresql-libpq/test/Test/Expr/OrderBy.hs
@@ -13,7 +13,7 @@ import qualified Orville.PostgreSQL.Internal.ExecutionResult as ExecResult
 import qualified Orville.PostgreSQL.Internal.Expr as Expr
 import qualified Orville.PostgreSQL.Internal.RawSql as RawSql
 
-import Test.Expr.TestSchema (FooBar (..), barColumn, dropAndRecreateTestTable, encodeFooBar, fooBarTable, fooColumn, insertFooBarSource, sqlValuesToText)
+import Test.Expr.TestSchema (FooBar (..), assertEqualSqlRows, barColumn, dropAndRecreateTestTable, encodeFooBar, fooBarTable, fooColumn, insertFooBarSource)
 import qualified Test.Property as Property
 
 orderByTests :: Pool.Pool Conn.Connection -> IO Bool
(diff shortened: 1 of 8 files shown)


orville-postgresql-libpq/test/Test/Expr/OrderBy.hs
```diff
@@ -87,4 +87,4 @@ runOrderByTest pool test =
 
           ExecResult.readRows result
 
-    rows HH.=== map encodeFooBar (orderByExpectedQueryResults test)
+    sqlValuesToText rows HH.=== sqlValuesToText (map encodeFooBar (orderByExpectedQueryResults test))
```
qxjit, line 90: I'd love to go a step further and have something like `assertEqualSqlRows` or something like that to hide the `sqlValuesToTextCalls`. That would also make it easier to change the implementation in the future if we wanted to.

orville-postgresql-libpq/test/Test/Expr/OrderBy.hs
```diff
@@ -87,4 +87,4 @@ runOrderByTest pool test =
 
           ExecResult.readRows result
 
-    sqlValuesToText rows HH.=== sqlValuesToText (map encodeFooBar (orderByExpectedQueryResults test))
+    rows `assertEqualSqlRows` map encodeFooBar (orderByExpectedQueryResults test)
```
qxjit, line 90: Any particular reason to use this function with infix notation at this point? All else being equal I'd opt for regular prefix function application here when switching away from the operator.

mitchellvitez: Just aesthetics. I think it reads better infixed since it's comparing a thing on each side just like `===`

Probably the best "reason" is that hedgehog calls these things `lhs` and `rhs`, but it shows the inner `===` line anyways

```haskell
        ┏━━ Test.hs ━━━
    212 ┃ testEq a b =
    213 ┃   sort a === sort b
        ┃   ^^^^^^^^^^^^^^^^^
        ┃   │ ━━━ Failed (- lhs) (+ rhs) ━━━
```

qxjit: Oh interesting, I wonder whether `withFrozenCallStack` would fix that.

qxjit (approved):

PR #165, opened 2021-08-04
Add orderBy and groupBy to SelectOptions

mitchellvitez:
Adds `orderBy` and `groupBy` to `SelectOptions`, as well as adding `(order|group)ByColumnsExpr` and a helper function for the `RawSql.commaSpace` separator


qxjit (approved):
:heart:

PR #166, opened 2021-08-09
Allow alternative timestamp field time formats in the database

OwenGraves:
This solution is kind of ugly, and there might be a better way to go about this (be happy to discuss), but this allows for parsing a few more representations in a timestamp field in the database (such as not having the milliseconds specified).


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/SqlValue.hs
```diff
@@ -251,7 +253,13 @@ toUTCTime sqlValue = do
...
   -- Leading to the ugly hacks below.
   txt <- toText sqlValue
-  Time.parseTimeM False Time.defaultTimeLocale "%F %T%Q%Z" (T.unpack txt <> "00")
+  let parseTime = Time.parseTimeM False Time.defaultTimeLocale
+      unTxt = T.unpack txt
+  -- We allow a couple alternative time representations in the database,
+  -- should probably expand this list or find a better way to do this
+  parseTime       "%F %T%Q%Z" (unTxt <> "00")
+    <|> parseTime "%F %T%Q"    unTxt
+    <|> parseTime "%F %T"      unTxt
```
qxjit, line 262: I was imagining that we would actually have a new `SqlType` to cover `timestamp without time zone`. Is that something you'd be willing to tackle rather than having this one parsing function try to handle both?

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/SqlType.hs
```diff
@@ -269,7 +271,25 @@ timestamp =
...
+
+  http://blog.untrod.com/2016/08/actually-understanding-timezones-in-postgresql.html
+-}
+timestampWithoutZone :: SqlType Time.UTCTime
+timestampWithoutZone =
+  SqlType
+    { sqlTypeExpr = Expr.timestampWithoutZone
+    , sqlTypeReferenceExpr = Nothing
+    , sqlTypeNullable = False
+    , sqlTypeId = LibPQ.Oid 1184
```
qxjit, line 289: I assume that `timestamp without time zone` has a different OID

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/SqlValue.hs
```diff
@@ -26,7 +26,8 @@ module Orville.PostgreSQL.Internal.SqlValue
     fromDay,
     toDay,
     fromUTCTime,
-    toUTCTime,
+    toUTCWithZoneTime,
+    toUTCWithoutZoneTime,
```
qxjit, line 30: Ugh.. naming these is hard. I think they should start with `toUTCTime` since they are creating a `UTCTime` value. Maybe `toUTCTimeWithZone` and `toUTCTimeWithoutZone`?

qxjit (changes requested):
The new `SqlType` and `FieldDefinition` should have at least the equivalent tests to those on the existing timestamp field.

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/SqlType.hs
OwenGraves: Yeah I was wondering about that, where do you look up OIDs?

telser: The easiest place is in postgres itself. A really rough query you can run is 
```sql
select typname, oid, pg_catalog.obj_description(oid) from pg_catalog.pg_type ;
```

Which should give you the name, the oid, and a description, but this isn't filtering out internal types or anything so the list is big (376 rows when I run it against the orville db container)

qxjit (approved):
:heart:

PR #167, opened 2021-08-09
LibPQ: Add `distinct`

mitchellvitez:
Adds `selectClause` and `selectExpr` under `queryExpr` to support `SELECT DISTINCT ...`. Doesn't support multiple-column `DISTINCT`s yet.

I'm a little unsure about the niceness of just passing a `Bool` to `selectExpr`. Can add a newtype wrapper to make it more clear what that flag means if wanted


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Select/SelectExpr.hs
```diff
@@ -0,0 +1,21 @@
...
+    selectExpr,
+  )
+where
+
+import qualified Orville.PostgreSQL.Internal.RawSql as RawSql
+
+newtype SelectExpr = SelectExpr RawSql.RawSql
+  deriving (RawSql.SqlExpression)
+
+selectExpr :: Bool -> SelectExpr
```
telser, line 19: I think you're right that a Bool feels sub-optimal at best. I think we should definitely explore something here, a newtype, or an ADT, whatever works best.

qxjit: Agree, Even a single constructor type `data Distinct = Distinct` would let us then write `Maybe Distinct -> Select Expr` here.

mitchellvitez: the grammar (see https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#set-quantifier) implies an ADT `data SetQuantifier = Distinct | All`, with `All` as a default. I do like having it be a `Maybe` with `Nothing` implying `All`, so we don't have to use the name `SetQuantifier` and the default feels more default

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/TableDefinition.hs
```diff
@@ -218,16 +219,20 @@ collectSetClauses entity fieldDef accessor clauses =
 -}
 mkQueryExpr ::
   TableDefinition key writeEntity readEntity ->
+  Maybe Expr.SelectClause ->
```
qxjit, line 222: It seems like this function always required an `Expr.SelectClause`, rather than leaving the user to wonder what happens if you don't specify one. Either than, or it should take a `Maybe Distinct`.

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/TableDefinition.hs
```diff
@@ -218,16 +219,20 @@ collectSetClauses entity fieldDef accessor clauses =
...
   Maybe Expr.GroupByClause ->
   Maybe Expr.LimitExpr ->
   Maybe Expr.OffsetExpr ->
   Expr.QueryExpr
-mkQueryExpr tableDef whereClause orderByClause groupByClause limitExpr offsetExpr =
+mkQueryExpr tableDef mbSelectClause whereClause orderByClause groupByClause limitExpr offsetExpr =
   let columns =
         marshallerColumnNames . tableMarshaller $ tableDef
+      selectClause =
+        fromMaybe (Expr.selectClause . Expr.selectExpr $ Just Expr.Distinct) mbSelectClause
```
qxjit, line 233: I wouldn't expect `DISTINCT` to be the default here when `Nothing` is passed (but see my comment above about not allowing `Nothing` anyway)

qxjit (changes requested):

qxjit (approved):

PR #168, opened 2021-08-11
[0.9] Add functions from Where/SelectOptions to (String, [SqlValue])

ysangkok:
These functions allow for sharing WhereConditions and SelectOptions between queries executed with higher-level primitives like `selectQueryTable` and lower-level ones like `selectQueryRawRows`.

The user is responsible for passing the obtained `SqlValue`s back into Orville in the right order.

The tests are failing because CircleCI doesn't seem to be configured for the 0.9 branch which this is for.


qxjit (approved):
:heart:

PR #170, opened 2021-08-17
Removes top-level foralls from Plan functions

qxjit:
This removes the `forall` on some top level plan functions to avoid
triggering the simplying-subsumption eta-expansion requirement in user
code on GHC 9.

I have left the `forall` on `PlanMany` to avoid letting a `Planned`
value containing a `Many` from escaping the plan scope. If such values
can escape escape the user can pass them directly to a new planning
scope, leading to very surprising results.


telser (approved):

aaronallen8455:
I'm curious if you have an example of code that fails to compile because of this.

aaronallen8455 (approved):

qxjit:
@aaronallen8455 Yes, it's on a branch in a private flipstone repo. I'll provide you the link offline.

PR #172, opened 2021-08-24
Adds libpq to haddock generation.

telser:
Also it looks like a cabal file was not generated correctly, so I've included that as well.


qxjit (approved):
:heart:

PR #174, opened 2021-08-30
Adds support for creating tables via auto-migration

qxjit:
This is a first baby step toward porting the auto migration capabilities
from the previous code base. For now, this only offers the ability to
create a table if it does not yet exist. It does not attempt to migrate
existing tables.


orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -0,0 +1,220 @@
...
+    MigrationDataError,
+  )
+where
+
+import Control.Exception (Exception, throwIO)
+import Control.Monad.IO.Class (liftIO)
+import Data.Foldable (traverse_)
+import Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Data.List.NonEmpty as NEL
+import qualified Data.Map.Strict as Map
```
telser, line 19: Maybe we should use a different qualifier for Strict/Lazy Map?

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Name/ColumnName.hs
```diff
@@ -17,5 +17,5 @@ newtype ColumnName
   = ColumnName RawSql.RawSql
   deriving (RawSql.SqlExpression)
 
-rawColumnName :: String -> ColumnName
-rawColumnName = ColumnName . RawSql.fromString
+columnName :: String -> ColumnName
+columnName = ColumnName . RawSql.fromString
```
telser, line 21: Do we want to make the distinction here for qualified/not? The nice thing about the raw prefix was it gave an indication that nothing was being done to transform the input. Escaping/qualification/etc. But maybe we should just do a full pass through on everything to handle that stuff consistently.

telser (approved):

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Name/ColumnName.hs
qxjit: @telser We now have `SqlExpression` for allowing you create a `ColumnName` from a raw sql without transformation. Once we add identifier quoting, this function would right apply it.

As far as qualification, I would treat that as a separate type, like we've done for `TableName`, but we don't have that case in the code yet.

orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
qxjit: @telser Perhaps. If so, I'd vote to make use `Map` for the strict one and then add something else for the lazy one (`LazyMap`?)

PR #190, opened 2021-09-15
Adds modules to query 'pg_catalog'

qxjit:
This adds modules for several 'pg_catalog' tables and a higher-level
'DatabaseDescription' for describing a number of tables at once using
'Plan'.

I had to augment 'Expr' an 'Plan' a bit to support new types, including
the ability to query using tuples. Although the 'Expr' module supports
any size tuple, it does not enforce that the caller using same-sized
tuples when building the queries.

In the higher order 'WhereCondition' and 'Operation' modules, I limited
the support to 2-tuples to easily enforce this without making the types
too crazy for the time being. We will want to iterate on this prior to
release.

For now, I stopped short of exposing any of this in 'Plan' directly.
Instead you currently have to reach down to 'Operation' to get the tuple
support.


orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
```diff
@@ -144,6 +151,104 @@ assertRight =
...
+data WherePlanner param = WherePlanner
+  { -- | The 'paramMarshaller' function provided here will be used to decode
+    -- the parameter field from the result set so that the row can be properly
+    -- associated with the input parameter that matched it.
+    paramMarshaller :: forall entity. (entity -> param) -> SqlMarshaller.SqlMarshaller entity param
+  , -- | 'executeOneWhereCondition' must build a where condition that will
+    -- match only those rows that match the input paramater.
+    executeOneWhereCondition :: param -> SelectOptions.WhereCondition
+  , -- | 'executeManyWhereCondition' must build a where condition that will
+    -- match only those rows that any (not all!) of the input parameters.
```
ysangkok, line 176: I expected a verb in final the subordinate clause. "That any of the input parameters" ... that the parameters what? That any of them do?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
```diff
@@ -144,6 +151,104 @@ assertRight =
...
+    executeManyWhereCondition :: NonEmpty param -> SelectOptions.WhereCondition
+  , -- | 'explainOneWhereCondition' must build a where condition that is
+    -- suitable to be returned as an explanation of
```
ysangkok, line 179: I think "as" is confusing here because it is unclear to me whether it is a role being played or an explanation. I suggest "because" ?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
```diff
@@ -144,6 +151,104 @@ assertRight =
...
+    -- 'executeManyWhereCondition' would return when given a parameter. For
+    -- example, this could fill in either an example or dummy value.
+    explainOneWhereCondition :: SelectOptions.WhereCondition
+  , -- | 'explainManyWhereCondition' must build a where condition that is
+    -- suitable to be returned as an explanation of
+    -- 'executeOneWhereCondition' would when given a list of parameters. For
```
ysangkok, line 185: executeManyWhereCondition on line 180 would "return". But here, what would executeOneWhereCondition do? There is no verb after "would". Is the intention to say that an explanation can be returned? But in comment block on executeManyWhereCondition, there are two occurances of "return". Why is there only one here?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
```diff
@@ -144,6 +151,104 @@ assertRight =
...
+    -- example, this could fill in either example or dummy values.
+    explainManyWhereCondition :: SelectOptions.WhereCondition
+  }
+
+{- |
+  Builds a 'WherePlanner' that will match on a single
+  'FieldDefinition.FieldDefinition'.  The resulting 'WherePlanner' can be used
+  with function such as 'findOne' and 'findAll' to construct an 'Operation'.
```
ysangkok, line 193: with *a* function?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
```diff
@@ -144,6 +151,104 @@ assertRight =
...
+        , executeOneWhereCondition = \fieldValue -> SelectOptions.fieldEquals fieldDef fieldValue
+        , executeManyWhereCondition = \fieldValues -> SelectOptions.fieldIn fieldDef fieldValues
+        , explainOneWhereCondition = SelectOptions.fieldEquals stringyField $ T.pack "EXAMPLE VALUE"
+        , explainManyWhereCondition = SelectOptions.fieldIn stringyField $ fmap T.pack ("EXAMPLE VALUE 1" :| ["EXAMPLE VALUE 2"])
+        }
+
+{- |
+  Builds a 'WherePlanner' that will match on a 2-tuple of
+  'FieldDefinition.FieldDefinition's.  The resulting 'WherePlanner' can be used
+  with function such as 'findOne' and 'findAll' to construct an 'Operation'.
```
ysangkok, line 212: with *a* function?

orville-postgresql-libpq/test/Test/PgCatalog.hs
```diff
@@ -0,0 +1,128 @@
...
+      Orville.findFirstEntityBy
+        PgCatalog.pgClassTable
+        (Orville.where_ $ Orville.fieldEquals PgCatalog.relationNameField pgClass)
+
+    fmap PgCatalog.pgClassRelationName result === Just pgClass
+    fmap PgCatalog.pgClassRelationKind result === Just PgCatalog.OrdinaryTable
+
+prop_queryPgAttribute :: Property.NamedDBProperty
+prop_queryPgAttribute =
+  Property.singletonNamedDBProperty "Can query the pg_attlibute table to find out about a column" $ \pool -> do
```
ysangkok, line 43: att*l*ibute should be attribute

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Where/BooleanExpr.hs
```diff
@@ -49,20 +54,57 @@ andExpr left right =
...
-      <> RawSql.intercalate RawSql.commaSpace (map RawSql.parameter $ NE.toList values)
-      <> RawSql.rightParen
+  inPredicate (columnReference columnName) (inValueList $ fmap valueExpression values)
 
 columnNotIn :: ColumnName -> NE.NonEmpty SqlValue -> BooleanExpr
 columnNotIn columnName values =
+  notInPredicate (columnReference columnName) (inValueList $ fmap valueExpression values)
+
+{- |
+  Checks that a the tuple constructed from the given columns in one of the
```
ysangkok, line 64: "a the"

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Where/BooleanExpr.hs
```diff
@@ -49,20 +54,57 @@ andExpr left right =
...
+  the tuples given have the same arity.
+-}
+columnTupleIn :: NE.NonEmpty ColumnName -> NE.NonEmpty (NE.NonEmpty SqlValue) -> BooleanExpr
+columnTupleIn columnNames valueLists =
+  inPredicate
+    (rowValueConstructor $ fmap columnReference columnNames)
+    (inValueList $ fmap (rowValueConstructor . fmap valueExpression) valueLists)
+
+{- |
+  Checks that a the tuple constructed from the given columns in NOT one of the
```
ysangkok, line 75: "a the". And "in NOT" should probably be "is NOT"

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/SqlValue.hs
```diff
@@ -147,11 +200,78 @@ toInt =
...
+
+{- |
+  Encodes an 'Word' value for usage with database
+-}
+fromWord :: Word -> SqlValue
+fromWord =
+  fromBSBuilderWithNoNULs BSB.wordDec
+
+{- |
+  Attempts to decode a 'SqlValue' a Haskell 'Word' value. If decoding fails
```
ysangkok, line 270: suspected missing preposition "to" or "as" before "a"

ysangkok (changes requested):
Grammar and spelling in comments 🧑‍🏫

orville-postgresql-libpq/test/Test/SqlType.hs
```diff
@@ -71,6 +72,40 @@ integerTests pool =
...
+    , runDecodingTest pool $
+        DecodingTest
+          { sqlTypeDDL = "SMALLINT"
+          , rawSqlValue = Just $ B8.pack $ show (32767 :: Int)
+          , sqlType = SqlType.smallInteger
+          , expectedValue = 32767
+          }
+    )
+  ,
+    ( String.fromString "Testing the decode of SMALLINT with value -2147483648"
```
telser, line 98: The value here is off

orville-postgresql-libpq/test/Test/SqlType.hs
```diff
@@ -71,6 +72,40 @@ integerTests pool =
...
+    , runDecodingTest pool $
+        DecodingTest
+          { sqlTypeDDL = "SMALLINT"
+          , rawSqlValue = Just $ B8.pack $ show (0 :: Int)
+          , sqlType = SqlType.smallInteger
+          , expectedValue = 0
+          }
+    )
+  ,
+    ( String.fromString "Testing the decode of SMALLINT with value 2147483647"
```
telser, line 88: Value off

orville-postgresql-libpq/src/Orville/PostgreSQL/PgCatalog/PgClass.hs
```diff
@@ -0,0 +1,145 @@
...
+{- |
+  A Haskell type for the name of the relation represented by a 'PgClass'
+-}
+newtype RelationName
+  = RelationName T.Text
+  deriving (Show, Eq, Ord, String.IsString)
+
+{- |
+  The kind of relation represented by a 'PgClass'
+-}
```
telser, line 47: Maybe it would be nice to include either here or on the to/from text functions a link to https://www.postgresql.org/docs/9.6/catalog-pg-class.html so future selves can find it?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan/Operation.hs
qxjit: I'll reword this. "because" doesn't make any sense here to me. The where condition referenced is literally "the explanation" -- i.e. the where clause that would be shown when doing asking for an explanation of a plan.

telser (approved):

Issue #189, opened 2021-09-14
Feature Request: Add support for Partial Indexes


martinezadrian:
It is sometimes useful to create a partial index that only applies to a subset of data. This seems achievable by just allowing the caller to provide an optional predicate on index creation.

https://www.postgresql.org/docs/8.0/indexes-partial.html

telser:
Supporting this is a very good idea. 

There are a few gotchas on the allowed expressions inside of the `WHERE` condition on partial indexes that we'll need to figure out. First is you cannot use a parameterized expression. The second being you can only operate on columns that are in the table you are creating the index against. This isn't super clear from the documentation until you read through the examples and find it under this section https://www.postgresql.org/docs/9.6/indexes-partial.html#INDEXES-PARTIAL-EX2. 

Adding this to the HDBC branch would need to probably either be unsafe in general or break backwards compatibility to get safety. But adding it to the libpq branch should be much more straightforward.

qxjit:
@telser Let's add this to HDBC in a simple-but-unsafe way to get us started. That would get is more familiar with whatever issues we would want to tackle with in for the libpq version.

PR #192, opened 2021-09-20
Adds functions to create partial indexes.

telser:
These are pretty unsafe and do no handholding whatsoever. Users need
to supply the conditions for the where clause as a list of strings.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/IndexDefinition.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/IndexDefinition.hs
@@ -6,6 +6,8 @@ License   : MIT
 module Database.Orville.PostgreSQL.Internal.IndexDefinition
   ( uniqueIndex
   , simpleIndex
+  , simplePartialIndex
+  , uniquePartialIndex
   ) where
 
 import Data.List (intercalate)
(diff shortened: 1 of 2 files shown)


telser:
This is to address #189

qxjit (approved):

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/IndexDefinition.hs
```diff
@@ -19,27 +21,55 @@ uniqueIndex ::
...
+mkIndexDefinition :: Bool
+                  -> String
+                  -> TableDefinition readEntity writeEntity key
+                  -> [SomeField]
+                  -> [String]
+                  -> IndexDefinition
+mkIndexDefinition unique name tableDef fields whereStrs =
+  let
+    whereStr [] = ""
+    whereStr strs = "WHERE " <> (intercalate " AND " strs)
```
ysangkok, line 68: what if the entries of whereStrs have an `OR` in them? the precedence rules would bind content from two different whereStrs entries together. this could be avoided by adding braces.

ysangkok (changes requested):

orville-postgresql/src/Database/Orville/PostgreSQL/Internal/IndexDefinition.hs
telser: I've updated to wrap each one in braces.

ysangkok (approved):
looks good, thanks!

PR #193, opened 2021-09-20
Adds AutoMigration of column types

qxjit:
This uses the `pg_attribute` information from `pg_catalog` to determine
whether the type of a column needs to altered, or nullability added or
removed.

I also revised `SqlType` a bit to match the way information is actually
used during migrations, including _NOT_ needing the length of integer
types to know whether they need to be migrate (the OID is sufficient).


orville-postgresql-libpq/src/Orville/PostgreSQL/PgCatalog/PgAttribute.hs
```diff
@@ -38,10 +40,49 @@ data PgAttribute = PgAttribute
...
+  or 'Nothing' if the length if the type is not variable.
+-}
+pgAttributeMaxLength :: PgAttribute -> Maybe Int32
+pgAttributeMaxLength attr =
+  -- This function is a port of the follow function from _pg_char_max_length function postgresql:
+  --
+  -- Note that it does not handle DOMAIN (user created) types correctly at
+  -- the moment. Doing so requires loading the pg_type record and checking
+  -- whether to use the attributes typid and typmod or domain types base type
+  -- and typemod when it is a domain type.
```
ysangkok, line 65: unclear precedence of and/or here

orville-postgresql-libpq/test/Test/AutoMigration.hs
```diff
@@ -98,8 +116,75 @@ prop_columnsWithSystemNameConflictsRaiseError =
...
-    (Orville.marshallField id (Orville.unboundedTextField "tableoid"))
+data SomeField where
+  SomeField :: Orville.FieldDefinition nullability a -> SomeField
+
+prop_altersColumnDataType :: Property.NamedDBProperty
+prop_altersColumnDataType =
+  Property.namedDBProperty "Alters data type on existing column" $ \pool -> do
+    let baseFieldDefs =
+          -- Serial columns are omitted from this list currently because
+          -- the are pseoud-types in postgresql, rather than real column types.
```
ysangkok, line 127: pseudo

orville-postgresql-libpq/test/Test/AutoMigration.hs
```diff
@@ -98,8 +116,75 @@ prop_columnsWithSystemNameConflictsRaiseError =
...
+          -- We don't handle migrating "away" from them to regular integer type.
+          --
+          -- time-like and boolean columns are omitted because postgresql raises
+          -- an unable-to-cast error when attempting to migrate between them an
```
ysangkok, line 131: s/an$/and/

ysangkok (changes requested):
comment nits

orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -178,18 +176,45 @@ mkAlterTableStep relationDesc tableDef =
...
+                    || (Orville.sqlTypeMaximumLength sqlType /= PgCatalog.pgAttributeMaxLength attr)
+
+                columnName =
+                  Orville.fieldColumnName fieldDef
+
+                dataType =
+                  Orville.sqlTypeExpr sqlType
+
+                alterType = do
+                  guard typeIsChanged
```
telser, line 203: I think this would be easier to follow with an `if` instead of the `guard`. But then I pretty much always prefer explicit branching in some form over the more hidden imperative forms.

qxjit: @telser I originally wrote both of these as an `if`, but thought it was too noisy. TBH, what I really wanted to do use `when`, but to do so collecting the migration steps would be a side effect in a write-like monad. I actually think that might be clearer overall because it would say something like

```
do
  when typeIsChanged (migrate $ alterColumnType...)
  when nullabilityIsChanged (migrate $ alterColumnNullability...)
```

I stopped short of actually adding a monad context to do that in, however. We can discuss further if we think it's worth it as we add more migration options.

ysangkok (approved):
looking good, thanks

orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
telser: I have an alternate proposal for that, but it can be discussed later when we get more migration options.

telser (approved):

PR #197, opened 2021-09-27
Fix statement parameter limit overflow for `insertRecordMany`

aaronallen8455:
If you attempt to run a statement with too many parameters, postgres
will throw an exception. We are seeing this occur with `insertRecordMany`.
There are fancier, more efficient ways to solve this issue (as shown in
[this blog post](https://klotzandrew.com/blog/postgres-passing-65535-parameter-limit)), but I have opted to use a simple batching strategy.

### DIFF
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Execute.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Execute.hs
@@ -13,7 +13,7 @@ import Database.Orville.PostgreSQL.Internal.Monad
 executingSql :: MonadOrville conn m => QueryType -> String -> IO a -> m a
 executingSql queryType sql action = do
   runningQuery <-
-    (\queryType -> ormEnvRunningQuery queryType)
+    (\queryTy -> ormEnvRunningQuery queryTy)
       <$> getOrvilleEnv
   liftIO $ runningQuery queryType sql (catchSqlErr sql action)
 
--- a/orville-postgresql/src/Database/Orville/PostgreSQL/Core.hs
+++ b/orville-postgresql/src/Database/Orville/PostgreSQL/Core.hs
@@ -361,21 +361,17 @@ insertRecordMany ::
 insertRecordMany tableDef newRecords = do
   let assignableColumnNames = tableAssignableColumnNames tableDef
       numParamsPerRow = length assignableColumnNames
-      -- Postgres has a hard limit of 65,536 parameters per statement since it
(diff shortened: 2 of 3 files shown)


orville-postgresql/src/Database/Orville/PostgreSQL/Internal/Execute.hs
```diff
@@ -13,7 +13,7 @@ import Database.Orville.PostgreSQL.Internal.Monad
 executingSql :: MonadOrville conn m => QueryType -> String -> IO a -> m a
 executingSql queryType sql action = do
   runningQuery <-
-    (\queryType -> ormEnvRunningQuery queryType)
+    (\queryTy -> ormEnvRunningQuery queryTy)
```
aaronallen8455, line 16: this was triggering a shadowed name warning

qxjit:
@aaronallen8455 How about we separate this out into an `insertBulkRecords` function that's dedicated to large datasets.

aaronallen8455:
@qxjit In short I think having two different bulk insert functions where the only difference is that one crashes if you give it a large dataset is rather confusing from a user's perspective. The updated function behaves the same if you use it with a small dataset. Are you concerned about the transaction?

aaronallen8455:
I'm in favor of having a bulk insert API, but having two different flavors (one of which is prone to crashing) is not ideal IMO.

qxjit:
@aaronallen8455 The reality is, there's a whole host of orville functions that could fail due to this some problem, not just inserts. `insertRecordMany` isn't special in that regard except perhaps it's the place you're likely to hit the problem _first_.  For libpq we'll look at supporting a real bulk API for inserts, but I don't think it's unreasonable to continue to supply `insertRecordMany` for the simple case, with documentation that leads to the bulk API when you need it.

aaronallen8455:
My argument for not continuing to provide it is that it clearly has a bug, one which caused a real headache for us in production. The patched version is functionally equivalent in the simple case and I would find it surprising if someone was so concerned about the potential for making additional queries that they want to risk it with the bugged version (the doc I added explicitly mentions this possibility).

qxjit:
@aaronallen8455 FWIW, I don't consider the current version to be "buggy". My interpretation is that the current function's behavior is to "insert many records at once in a single insert statement". No matter what we do, that single insert statement could always hit Postgresql's limit, even if we did some internal optimizations to push that limit out further. A function that achieves bulk insert via _multiple_ insert statements is a different behavior, so I would like to offer that via a different function.

aaronallen8455:
We ended up working around this in our application code.

PR #200, opened 2021-10-01
Upgrade Stack to 2.7.1

ysangkok:
Projects that use Orville by way of submodule, and use a Stack version newer than 1.9.3, may also have a newer hpack version, which upon execution will then overwrite the old cabal file and cause the submodule's tree to be dirty. By upgrading to a Stack release that consumers are more likely to use, this nuisance is minimized.

Because the generated Cabal file only contains cosmetic changes, builds made by Cabal will be unaffected.

### DIFF
--- a/orville-postgresql/Dockerfile
+++ b/orville-postgresql/Dockerfile
@@ -1,4 +1,4 @@
-FROM flipstone/stack:v2-1.9.3
+FROM flipstone/stack:v4-2.7.1
 
 # The fpco.list apt source was causing an error build:
 #   E: The method driver /usr/lib/apt/methods/https could not be found.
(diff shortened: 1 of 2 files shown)


qxjit:
@ysangkok I think it's ok to merge this to work around your current issue, but I think the better long term answer here is either

1) not use submodules because `stack` does this
2) find a way to tell `stack` to use the cabal file rather than the `package.yaml`
3) only have a `.cabal` file in orville in the first place.

Having to keep orville's version of `stack` in sync with the projects using as an extra-dep from github is not something I think we can support long term.

(Of course, eventually we'll get it on hackage and the problem with be mostly gone anyhow)

qxjit (approved):

PR #203, opened 2021-10-12
Adds support for unique constraints

qxjit:
This adds support for unique constraints at the table level. Constraints
are added and dropped from the database during auto migration based on
whether an equivalent constraint exists.

This logic will be quickly improved by the following work, which will
add support for foreign key constraints, thus requiring support for two
different kinds of constraints.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
@@ -12,12 +12,14 @@ module Orville.PostgreSQL.AutoMigration
   )
 where
 
+import Control.Arrow ((&&&))
 import Control.Exception.Safe (Exception, throwIO)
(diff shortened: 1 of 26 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -174,9 +176,30 @@ mkAlterTableStep relationDesc tableDef =
...
+      existingConstraints =
+        Set.fromList
+          . Maybe.mapMaybe (pgConstraintMigrationKey relationDesc)
+          . PgCatalog.relationConstraints
+          $ relationDesc
+
+      constraintsToKeep =
+        Map.keysSet
+          . Orville.tableConstraints
+          $ tableDef
```
jbrechtel, line 189: Is this `$` and the one above it just here to align with the `.` ?

PR #204, opened 2021-10-17
Adds support for Foreign Key Constraints

qxjit:
This adds support for adding foreign keys as table constraints to the
databases schema by allowing foreign key `ConstraintDefinition`s to be
built and added to a `TableDefinition`.

This required 2 main changes in the `pg_catalog` and auto-migration
handling.

1) I added `ConstraintDescription` to load the details of the foreign
   relation and columns of the foreign key so that auto-migration can
   detect whether the foreign key exists. This now resolves the
   atttribute numbers from `PgConstraint` to `PgAttribute` records in
   the `DatabaseDescription` itself rather than requiring the
   auto-migration logic to lookup the attribute names later. If any
   of the attribute numbers can't be successfully looked up loading the
   entire database description will fail.

2) Creating and altering tables is now done in multiple steps to ensure
   that the columns and required unique indexes exist before the foreign
   keys are are added, and that foreign keys are dropped. Rather doing
   dependency analysis to determine a minimal number of migration steps
   I simply annotated the mirgation steps with a type so that can be
   sorted into an order that will succeed.


orville-postgresql-libpq/test/Main.hs
```diff
@@ -1,10 +1,12 @@
 module Main
   ( main,
+    recheckDBProperty,
```
telser, line 3: I imagine this is hugely useful!

PR #216, opened 2022-01-19
HDBC: Adds an `insertRecordManyReturning` function

jlavelle:
Also rexpressed `insertRecord` in terms of the new function and added a
test case.


qxjit:
@jlavelle This change should be safe to make on the `Orville-0.9` since it's a non-breaking change. Would you mind making on that branch first, and then merging `Orville-0.9` back to `master` get the changes there?

PR #219, opened 2022-01-26
LibPQ: Adds error information to MarshallError

qxjit:
This adds detail error information to the `MarshallError` produced by
`SqlMarshaller` when marshalling fails. This includes row identifiers
for the row being decode, as well as details about the column and values
being decode and any more general message provided by the source of the
decoding failure.

A new `ErrorDetailLevel` type controls the level of detailed actually
included in when the `MarshallError` is rendered to a `String` to be
displayed. This can be specified manually simply by catching the error
and providing the desired detail level to the render function. A default
value can be provided as well when constructing the `OrvilleState`,
which is what will be used if the `MarshallError` is rendered as an
`Exception` (via its `Show` instance).

The `SqlMarshaller` lacks any notion of which columns might be the row
identifiers to inlude in the error message. A new wrapper type, the
`AnnotatedSqlMarshaller` now captures that information. This new type
is automatically created when the `TableDefinition` is built, using
the primary key of the table as the row id.

### DIFF
--- a/orville-postgresql-libpq/orville-postgresql-libpq.cabal
+++ b/orville-postgresql-libpq/orville-postgresql-libpq.cabal
@@ -33,9 +33,11 @@ library
       Orville.PostgreSQL.AutoMigration
       Orville.PostgreSQL.Connection
       Orville.PostgreSQL.Internal.DefaultValue
+      Orville.PostgreSQL.Internal.ErrorDetailLevel
       Orville.PostgreSQL.Internal.ExecutionResult
       Orville.PostgreSQL.Internal.Expr
       Orville.PostgreSQL.Internal.FieldDefinition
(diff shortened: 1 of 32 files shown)


OwenGraves (approved):
Nice! This will be helpful.

Are Strings (over Text) used throughout for an interesting reason, or just mostly to be compatible with existing code?

qxjit:
@OwenGraves Since the primary usage of the rendering functions is for the `Show` and `Exception` instances of `MarshallError`, it was simplest to use `String` throughout them rather than use `Text` only then need to convert it to `String` later. Error messages in general (and these in particular) tend not to be very large, so the performance implications of using `String` are at least mitigated by the amount of text involved.

telser (approved):

PR #220, opened 2022-02-16
Adds `WhereCondition` operators

qxjit:
This adds a number of operators for building and combining
`WhereCondition` values for those that would like to use use them.

In adding these operators I realized we had missed creating binary
boolean functions for `AND` and `OR`. I ended up deciding `whereAnd` and
`whereOr` should simply be those functions and that the existing
versions of the functions were not really necessary. A user can also do
a fold using `whereAnd` or `whereOr` as desired if they have a dynamic
list of conditions. For literals, the operators are available to make
this nice be, such as

```
      (fieldA .== a)
  .&& (fieldB .== b)
  .&& (fieldB .== b)
```

Or, alternatively

```
    whereAnd (fieldA .== a)
  . whereAnd (fieldB .== b)
  $          (fieldB .== b)
```


orville-postgresql-libpq/test/Test/SelectOptions.hs
```diff
@@ -184,21 +296,25 @@ prop_groupByCombined =
           <> (SO.groupBy . Expr.groupByExpr . RawSql.toRawSql $ FieldDef.fieldColumnName barField)
       )
 
-assertDistinctEquals :: HH.MonadTest m => String -> SO.SelectOptions -> m ()
+assertDistinctEquals :: (HH.MonadTest m, HasCallStack) => String -> SO.SelectOptions -> m ()
 assertDistinctEquals mbDistinct selectOptions =
-  RawSql.toExampleBytes (SO.selectDistinct selectOptions) HH.=== B8.pack mbDistinct
+  withFrozenCallStack $
```
telser, line 301: Why do we have this?

qxjit: So that when the assertion fails you see where the assertion fails in a test you see where the assertion was called from rather than the definition of the assertion itself.

telser: Ah, I figured it was something like that! Awesome, thanks!

telser (approved):

PR #224, opened 2022-02-28
Ports transaction callbacks to libpq-Orville

qxjit:
Transaction callbacks allow a user to add an IO operation that will be
called after Orville runs a transaction-related action like beginning or
commiting a transaction.


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Transaction.hs
```diff
@@ -47,44 +47,48 @@ withTransaction action =
...
 
     committed <- liftIO $ IORef.newIORef False
+    callback <- OrvilleState.orvilleTransactionCallback <$> OrvilleState.askOrvilleState
+
     let doAction = do
-          liftIO $ RawSql.executeVoid conn (openTransactionSql transaction)
+          liftIO $ do
+            let openEvent = OrvilleState.openTransactionEvent transaction
+            RawSql.executeVoid conn (transactionEventSql openEvent)
+            callback openEvent
```
aaronallen8455, line 56: The API documentation should probably specify that if the callback throws then the transaction will fail. Specifically, the user supplied action needs to handle exceptions if they don't want this behavior.

qxjit: Yes, that's a good point. I'll add information about that to docs.

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Transaction.hs
```diff
@@ -47,44 +47,48 @@ withTransaction action =
...
           value <-
             OrvilleState.localOrvilleState
               (OrvilleState.connectState innerConnectedState)
               action
           liftIO $ do
-            RawSql.executeVoid conn (transactionSuccessSql transaction)
+            let successEvent = OrvilleState.transactionSuccessEvent transaction
+            RawSql.executeVoid conn (transactionEventSql successEvent)
             liftIO $ IORef.writeIORef committed True
+            callback successEvent
```
jbrechtel, line 66: On the topic of docs and exceptions - if I'm reading this right then an exception in the open transaction callback will cause the transaction to rollback but an exception in the success callback will not. Seems like that's probably I'd normally want and expect but maybe annoying to describe.

jbrechtel (approved):
I see a few typos in the docs for `withTransaction` on 28 `comitted` is missing an m and 31 "This function in save to call " s/in save/is safe

telser (approved):

aaronallen8455 (approved):

PR #228, opened 2022-05-13
Adds support for nulls first and last ordering

jlavelle:
Ordering with nulls first or last is now supported via the new
`ascendingOrderWith` and `descendingOrderWith` functions. These both
accept an argument of type `NullsOrder` to specify the null ordering.


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/OrderBy/OrderByDirection.hs
```diff
@@ -17,8 +20,26 @@ import qualified Orville.PostgreSQL.Internal.RawSql as RawSql
...
+  | NullsLast
+  deriving (Eq, Show, Ord, Enum, Bounded)
+
 ascendingOrder :: OrderByDirection
 ascendingOrder = OrderByDirection $ RawSql.fromString "ASC"
 
 descendingOrder :: OrderByDirection
 descendingOrder = OrderByDirection $ RawSql.fromString "DESC"
+
+ascendingOrderWith :: NullsOrder -> OrderByDirection
```
qxjit, line 34: :thinking: Perhaps `ascendingOrder` and `descendingOrder` should take a `Maybe NullsOrder` parameter rather than needing separate functions?

jlavelle: I thought about this but wanted to maintain backwards compatibility, is that a concern?

jlavelle: Also if we go with taking a `Maybe NullsOrder` then you always have to write `ascendingOrder Nothing` in when you just want the default ordering.

qxjit: Ah, that's a good point -- these are exposed all the way up through `SelectOptions` as public API and having to specify `Nothing` on every ordering would probably get annoying when working at the higher level.

qxjit: Given that, we should probably export these new functions all the places that `ascendingOrder` is exported right now.

jlavelle: Alright, I think I got all of the exports with [c936d83](https://github.com/flipstone/orville/pull/228/commits/c936d83a92aca9a3949797b00ccd689b9b76456c)

qxjit (approved):

PR #229, opened 2022-05-17
Ports orville 0.9 triggers to orville-libpq as EntityTrace

qxjit:
This ports the 'Database.Orville.PostgreSQL.Internal.Trigger' module
from orville 0.9 to libpq as the 'Orville.PostgreSQL.EntityTrace'
module. This is renamed to avoid any confusion that it may be related to
database triggers.

The port involves the following structure changes:

* The trace callbacks are addeded to a `TracedTable` entity rather than
  being implemented via typeclasses
* The transaction state management support savepoints (since libpq
  orville support savepoints for nested transactions).
* `EntityTraceT` is a type-synonym for a `ReaderT` type rather than a
  newtype, which keeps with the pattern of orville libpq not providing
  it's own custom monad transformer newtypes.
* `MonadEntityTrace` includes a `liftOrvilleUntraced` member, allowing
  `insertEntityTraced` et al to not required a `MonadOrville` instance
  directly. This allows a user the option of building a monad that
  implements _only_ `MonadEntityTrace` if they wish to further enforce
  that the `*Traced` functions are used rather than the regular orville
  operations.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
@@ -1,0 +1,710 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+
+module Orville.PostgreSQL.EntityTrace
+  ( EntityTraceT,
+    EntityTraceState,
+    runEntityTraceT,
+    insertEntityTraced,
+    updateEntityTraced,
+    deleteEntityTraced,
+    TracedTable,
+    mkTracedTable,
+    addInsertTrace,
+    addUpdateTrace,
+    addDeleteTrace,
+    untracedTableDefinition,
+    MonadEntityTrace (recordTraces, liftOrvilleUntraced),
+  )
+where
+
+import qualified Control.Exception as Ex
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Control.Monad.Trans.Class (MonadTrans (lift))
+import Control.Monad.Trans.Reader (ReaderT, ask, runReaderT)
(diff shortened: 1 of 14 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
```diff
@@ -0,0 +1,710 @@
...
+  Constructs a 'TracedTable' that wraps the given 'O.TableDefinition' so that
+  it can be used with functions such as 'insertEntityTraced'. The resulting
+  'TracedTable' has no trace callbacks initially. You can 'addInsertTrace' and
+  similar functions ta add callbacks.
+-}
+mkTracedTable ::
+  (readEntity -> key) ->
+  O.TableDefinition (O.HasKey key) writeEntity readEntity ->
+  TracedTable trace key writeEntity readEntity
+mkTracedTable getEntityKey tableDef =
```
jbrechtel, line 90: I sort of wish we had a way to have our cake and eat it to so we could put the traces in the call to `mkTracedTable` with named fields without having to expose the record constructor too. As it is we'll have to stack our tracing functions in-front of `mkTracedTable` ala


```
addDeleteTrace ... $
addInsertTrace ... $
mkTracedTable ...
```

But honestly it's a pretty minor ergonomic nit and I don't have a suggestion I like for changing that.

jbrechtel (approved):
:+1: Great! This looks like exactly what I had in mind per our discussions.

The handling of transactions and savepoints is clear enough to follow - despite it being somewhat involved since it's dealing with a dicey problem. That's nice.

orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
qxjit: @jbrechtel We could have a separate record that we expose like we have for `TableParams` with the Orville 0.9 table definition, but TBH I don't love that. You also have the option of using [&](https://hackage.haskell.org/package/base-4.16.1.0/docs/Data-Function.html#v:-38-) to line up the traces afterwards, e.g.

```
myTracedTable =
  mkTraceTable keyAcessor someTableDef
     & addInsertTrace insertTraceFunc
     & addDeleteTrace deleteTraceFunc
```

PR #232, opened 2022-06-06
Ports updateFields to LibPQ

qxjit:


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Update.hs
```diff
@@ -75,26 +79,76 @@ updateToTableReturning ::
...
+  [Expr.SetClause] ->
+  Maybe SelectOptions.WhereCondition ->
+  Update readEntity ReturningClause
+updateToTableFieldsReturning =
+  updateFields WithReturning
+
+updateFields ::
+  ReturningOption returningClause ->
+  TableDefinition key writeEntity readEntity ->
+  [Expr.SetClause] ->
```
aaronallen8455, line 127: Does an empty list produce invalid SQL? Consider using `NonEmpty` here.

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/TableDefinition.hs
```diff
@@ -291,11 +291,15 @@ mkTablePrimaryKeyExpr tableDef =
     TableHasNoKey ->
       Nothing
 
-mkReturningClause ::
+{- |
+  When 'WithReturning' is given, builds a 'Expr.ReturningExpr' that will
+  return all the columns in the give table definition.
```
aaronallen8455, line 296: give -> given

aaronallen8455 (approved):

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Update.hs
qxjit: Good point. I may have ported Orville 0.9 a little too faithfully in this case

PR #235, opened 2022-06-20
Adds ability to customize begin transaction sql

qxjit:
In order to test this I ensured that the transaction sql statements made
it into the sql executino callbacks so I could trace them in the test.
This required providing execute functions that both delivery callbacks
and work in IO so that the transaction callback could be made inside the
same `liftIO` as the sql execution.

When adding the test for this I discovered that our `beginTransaction`
expression builder built incorrect sql when used with an isolation
level. I fixed that as well.


orville-postgresql-libpq/test/Test/Transaction.hs
```diff
@@ -127,6 +132,34 @@ prop_callbacksMadeForTransactionRollback =
...
+          [ Expr.beginTransaction Nothing
+          , Expr.beginTransaction (Just Expr.readOnly)
+          , Expr.beginTransaction (Just Expr.readWrite)
+          , Expr.beginTransaction (Just Expr.deferrable)
+          , Expr.beginTransaction (Just Expr.notDeferrable)
+          , Expr.beginTransaction (Just (Expr.isolationLevel Expr.serializable))
+          , Expr.beginTransaction (Just (Expr.isolationLevel Expr.repeatableRead))
+          , Expr.beginTransaction (Just (Expr.isolationLevel Expr.readCommitted))
+          , Expr.beginTransaction (Just (Expr.isolationLevel Expr.readUncommitted))
+          ]
```
aaronallen8455, line 150: It'd be better to do the check once for each item in the list rather than randomly selecting one `n` times.

aaronallen8455 (approved):

PR #237, opened 2022-07-07
Implement a Date parser in attoparsec

cmika:
The date parser has a small amount of leniency when it comes to
parsing years, the ISO spec states that years should not be less
than 4 digits and the parser allows for smaller years. Postgres
should not give us a date like that, and adding support for that
would mean reimplementing the `decimal` function, so I chose to
allow the leniency.

I added a couple more cases for Date tests.


telser:
For future reference we'll probably want to look at supporting the different styles of date/time output from postgres which would break this. https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-OUTPUT

PR #238, opened 2022-07-12
Use Attoparsec to parse local time and utc time

cmika:
This adds parsers for localTime and utcTime along with some internal
helpers for smaller pieces.

I chose to make the seconds parser parse any size seconds fraction
component, and it will truncate if it doesn't fit into the picosecond
resolution that Data.Time.TimeOfDay uses.

During implementation of seconds I had to implement a alternative decimal,
so I refactored date parsing to use this and it now rejects years less
than 4 digits long as perscribed by the ISO 8601.

Profiling my project with this reduced allocations from 3.8 gigs to 2.7 gigs for our benchmark.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/PgTime.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/PgTime.hs
@@ -8,8 +8,8 @@ module Orville.PostgreSQL.Internal.PgTime
   )
 where
 
-import qualified Data.Attoparsec.ByteString.Char8 as AttoB8
 import qualified Data.Attoparsec.ByteString as AttoBS
+import qualified Data.Attoparsec.ByteString.Char8 as AttoB8
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.Char as Char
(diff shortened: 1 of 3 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/PgTime.hs
```diff
@@ -29,21 +30,25 @@ dayToPostgreSQL =
...
-twoDigits :: AttoB8.Parser Int
+twoDigits :: Integral a => AttoB8.Parser a
 twoDigits = do
   tens <- AttoB8.digit
   ones <- AttoB8.digit
   pure $ fromChar tens * 10 + fromChar ones
   where
-    fromChar c = Char.ord c - Char.ord '0'
+    fromChar :: Integral a => Char -> a
+    fromChar c = fromIntegral $ Char.ord c - Char.ord '0'
```
qxjit, line 51: Let's make `fromChar` a top-level function rather than `do ... where`

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/PgTime.hs
```diff
@@ -78,37 +82,45 @@ localTimeToPostgreSQL =
...
+  An Attoparsec parser for parsing a base 10 number and returns the number of
+  digits consumed. Based off of AttoB8.decimal.
+-}
+decimalWithCount :: Integral a => AttoB8.Parser (a, a)
+decimalWithCount = do
+  wrds <- AttoBS.takeWhile1 AttoB8.isDigit_w8
+  pure (BS.foldl' step 0 wrds, fromIntegral $ BS.length wrds)
+  where
+    step :: Integral a => a -> Word.Word8 -> a
+    step a w = a * 10 + fromIntegral (w - 48)
```
qxjit, line 113: Let's make `step` a top-level function instead of `do ... where`

qxjit (approved):

PR #240, opened 2022-07-30
Exposes the `Delete`, `Insert`, and `Update` modules

jlavelle:
I also exposed the `raw*Expr` functions from each of the modules.


qxjit:
@jlavelle These updates should be made in the `package.yaml` file that then generates the `.cabal` file.

jlavelle:
@qxjit oops, I totally forgot about hpack :woman_facepalming:. Fixed.

qxjit (approved):

PR #241, opened 2022-08-02
LibPQ - Adds Expr definitions for Cursor statements

qxjit:
These cursor statements will be used to build a higher level API to
interact with cursor and stream results out of the database.


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Cursor.hs
```diff
@@ -0,0 +1,232 @@
...
+
+absolute :: Int -> CursorDirection
+absolute countParam =
+  -- postgresql won't let us pass the count as a parameter.
+  -- when we try we get an error like such error:
+  --  ERROR:  syntax error at or near "$1"
+  --  LINE 1: FETCH ABSOLUTE $1 \"testcursor\"
+  CursorDirection $
+    RawSql.fromString "ABSOLUTE "
+      <> RawSql.fromString (show countParam)
```
aaronallen8455, line 172: Could define a `RawSql.fromInt` helper function

aaronallen8455 (approved):
I'd suggest exploring `PQsetSingleRowMode` as a way to back streaming functionality. It's supposedly more efficient than using a cursor and might make for a simpler implementation as well.

telser (approved):
While single row mode could be better than this in some (many?) situations. I think there are enough potential gotchas to at least explore that as a future enhancement.

PR #258, opened 2023-01-24
Adds Expr expressions for some time and math related functions.

qxjit:
This provides enough time and math expression to allow the user to write expressions such as `now() - interval '1 hour'` and the similar expressions. It also includes a number of other binary math operators as well since they were trivial to add.

This change introduces more parentheses into Orville's generated SQL because the inclusion of math operators seemed to make it appropriate to make a number of Expr functions that work with `ValueExpression` to add their parentheses around their arguments to avoid ambiguity. This allowed some parentheses to be removed from more abstract functions (such as `fieldEquals`), but results in more parentheses overall.


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/DataType.hs
```diff
@@ -0,0 +1,113 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{- |
+Module    : Orville.PostgreSQL.Expr.ColumnDefinition
```
telser, line 4: The haddock here is wrong

telser (approved):

orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/DataType.hs
qxjit: Thank you!

PR #260, opened 2023-02-18
Create GETTING-STARTED.md

ysangkok:
I have been searching for tools that allow for testable documentation, and this is what I came up with so far. It's finicky (especially the list-bin part), but submitting it here for review in case anybody has suggestions on how to improve it.

### DIFF
--- a/GETTING-STARTED.md
+++ b/GETTING-STARTED.md
@@ -1,0 +1,96 @@
+#!/usr/bin/env mdsh
+
+# Getting started with Orville
+
+Orville is a PostgreSQL client library, so it needs a working server to work. The following code block starts a server and makes a user for us to use. Your operating system probably has a custom way to do this, so consult their documentation.
+
+```shell
+screen -d -m su -c '/usr/lib/postgresql/11/bin/postmaster -D /etc/postgresql/11/main/' postgres
+sleep 5
+cat << EOF > create-user
+create database root;
+create user root;
+grant all privileges on database root to root;
+EOF
+su -c 'psql -f create-user' postgres
+```
+
+For this demo, let's make a new project:
+
+```shell
+mkdir orville-getting-started
+cd orville-getting-started
+cabal init -n --exe
+```
+
+Make it depend on Orville (the LibPQ variant, which is preferred!), resource-pool and bytestring, which Orville integrates with:
+
+```shell
+sed -i -re 's/build-depends:/build-depends: orville-postgresql-libpq, bytestring, resource-pool,/' *.cabal
+```
+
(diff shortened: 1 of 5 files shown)


qxjit:
@ysangkok This is great, thank you! Just a couple things I'd like to have before merging this

1) Can we make the getting started example not use any language extensions please?
2) Can we get a dockerized setup of `mdsh` that is run as port of `scripts/build.sh` and ci to ensure that we keep this example working

GETTING-STARTED.md
```diff
@@ -0,0 +1,82 @@
...
+cat <<EOF > readme-test.t
+$ cd \$TESTDIR
+$ cp \$(cabal list-bin exe:\$(basename \$PWD) | tail -n1) \$OLDPWD
+$ cd \$OLDPWD
+$ ./orville-demo-*
+Right 2
+EOF
+```
+
+It can be verified to match using `prysk` (which can be installed using `pip3`):
```
telser, line 78: Maybe we should have a beginning section describing that this is based on Debian (and a version?) along with installing prerequisites and noting what they are for? That way we could level set at the very beginning. 

I didn't know of `prysk`, but it looks interesting.

ysangkok: It's just a forked https://bitheap.org/cram/

ysangkok: I added the Docker setup David suggested, does that take care of your concern?

ysangkok:
@qxjit Added a Docker setup as suggested

qxjit:
@ysangkok @telser  I tried the new `build.sh` locally and it doesn't seem to be building the `orville-docs` image for some reason. I think something is wrong with the `SKIP_IMAGE_BUILD` conditional or variable. It doesn't do a docker build whether I pass `--skip-image-build` or not.

If I remove the conditional wrapper then the `cd ../../docs` fails. It looks like it has one too many `..`s in it.

ysangkok:
Where is `build.sh` supposed to be called from? Should it be agnostic to the working directory? I call it from the directory where it is, but it looks like you'd expect it to work when invoked from a different directory?

qxjit:
@ysangkok We've been calling it from the `orville-postgresql-libpq` by using `./scripts/build.sh`.

ysangkok:
Ok, I changed to support that. I avoided trying to work from any directory since it seems only possible with bashisms: https://stackoverflow.com/a/246128/309483

qxjit:
@ysangkok I fixed the issues with the build script not building the docker images. This is good to merge with that commit included

qxjit (approved):

PR #261, opened 2023-02-19
Removes wrapper types from SelectOptions fields

qxjit:
This removes the `OrderBy` and `WhereCondition` types which had become
just newtype wrappers around `OrderByExpr` and `BooleanExpr`
respectively. Removing these wrappers allows users to make a smoother
transition from working with `SelectOptions` to building `Expr`
expressions directly because the helpers the are used to using
(`orderByField`, `fieldEquals`, etc) can be still be used.


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/Where/BooleanExpr.hs
```diff
@@ -51,13 +53,29 @@ orExpr left right =
     (booleanValueExpression left)
     (booleanValueExpression right)
 
+{- |
+  Operator alias for 'orExpr'
+-}
+(.||) :: BooleanExpr -> BooleanExpr -> BooleanExpr
+(.||) = orExpr
+
+infixr 8 .||
```
telser, line 62: This is a bit tricky I think. To begin on both the Haskell side and PostgreSQL side logical OR and AND have different precedence. However, they appear to be defined with _opposite_ fixities. According to the PostgreSQL docs they associate on the left, as opposed to the right binding here (and in base for both `||` and `&&`).  To complicate things even further, they seem to have opposite precedence.

If I'm a user, would I expect this to be more like writing Haskell, or to have a more direct translation to SQL? I can imagine some subset of users being confused either direction. 

[base &&/||](https://www.stackage.org/haddock/lts-20.11/base-4.16.4.0/Prelude.html#v:-38--38-)

[PostgreSQL operator precedence](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE)

qxjit: @telser That's a good point. Let's discuss that outside the scope of this PR since this isn't change the precedence of fixity of those operators, only moving them and changing their argument types.

qxjit: @telser Also, we always put parens around the SQL expressions given to and/or, so any difference in precedence between Haskell and Postgres shouldn't affect the boolean logic accidentally.

telser: Oh yes I wasn't thinking about accidental logic changes on Orville's part, but entirely about the user expectation. I'm happy to talk about that separately though!

telser (approved):
I do wonder if there is a loss of clarity with this change, but the utility is extremely clear to me. Given that one is a wonder and the other is certain I overall like the change.

qxjit:
@telser I think any loss of clarity will become clear when we start organizing the exports more carefully and examining the user experience in the docs. If we find anything there we'll try to work within the basic setup in this PR (i.e. not re-introducing the wrappers) to make it more clear.

PR #264, opened 2023-02-24
Add migration tutorial

ysangkok:
Also adding an export for the DataType constructor, since it is needed for JSON support, which I will add a tutorial for in an upcoming PR.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/DataType.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/DataType.hs
@@ -6,7 +6,7 @@ Copyright : Flipstone Technology Partners 2016-2023
 License   : MIT
 -}
 module Orville.PostgreSQL.Internal.Expr.DataType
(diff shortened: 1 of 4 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Internal/Expr/DataType.hs
```diff
@@ -6,7 +6,7 @@ Copyright : Flipstone Technology Partners 2016-2023
 License   : MIT
 -}
 module Orville.PostgreSQL.Internal.Expr.DataType
-  ( DataType,
+  ( DataType (..),
```
qxjit, line 9: Our intention is not to export the constructors of the various expression types but instead direct users to use the `SqlExpression` class api when they need to construct new primitive values of the types.

ysangkok: Fixed in 82721397bee53c2e0aad84c9c4c24953224d652e

PR #265, opened 2023-02-27
Add JSON tutorial

ysangkok:
I decided that JSON deserves its own document. Will cover the timezone caveats in a separate document.

### DIFF
--- a/JSON.md
+++ b/JSON.md
@@ -1,0 +1,228 @@
+#!/usr/bin/env mdsh
+
+# Using JSON
+
+SQL has a rigid schema. Using JSON inside SQL allows for pockets of schemaless
+data, that is still queryable using PostgreSQL's built-in functionality.
+
+This document explains how to use Orville with the JSONB data type that
+PostgreSQL natively supports. The data type isn't built in to Orville, but it's
+shown here how to add it.
+
+Project initialization is similar to previous tutorials, but additional
+dependencies like Aeson have been added. Aeson is a JSON library for Haskell.
+
+```shell
+mkdir orville-json
+cd orville-json
+cabal init -n --exe
+sed -i -re 's/build-depends:/build-depends: orville-postgresql-libpq, aeson, bytestring, postgresql-libpq, resource-pool, text, vector,/' *.cabal
+cat << 'EOF' > cabal.project
+packages: .
+constraints: resource-pool < 0.3
+source-repository-package
+  type: git
+  location: https://github.com/flipstone/orville.git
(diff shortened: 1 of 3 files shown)


JSON.md
```diff
@@ -0,0 +1,228 @@
...
+nulls.
+
+We could also use a custom type with `FromJSON`/`ToJSON` instances, since
+`jsonb` allows for that too. Aeson is not the focus of this document though.
+
+```shell
+cat << 'EOF' >> app/Main.hs
+data Foo = Foo
+  { fooId :: Int.Int32
+  , fooTags :: Value
```
qxjit, line 68: I imagine we'll update this tutorial to use `json-fleece` eventually

JSON.md
```diff
@@ -0,0 +1,228 @@
...
+cat << 'EOF' >> app/Main.hs
+    result <- RawSql.execute connection (RawSql.fromString "SELECT id, jsonb_array_elements(tags) AS tag FROM json_demo")
+EOF
+```
+
+We can fetch the results with the low-level `readRows` API:
+
+```shell
+cat << 'EOF' >> app/Main.hs
+    [[_, (_, one)], [_, (_, two)], [_, (_, three)]] <- O.readRows result
```
qxjit, line 185: Maybe we should use `executeAndDecode` with a mashaller rather than `readRows` here?

I'm a little concerned about tutorials that show a quick way to do something that is never how we would do it in real code

ysangkok: I was aiming to show it both using a low and a high-level method. If this block uses a marshaller, then what would the `marshallResultFromSqlUsingRowIdExtractor` block below show? Not sure what to do here. If you think it is best, I can just remove the use of `readRows`?

JSON.md
ysangkok: Sure, if you think it is ready now, I can add it. But I also think it might be overwhelming for a potential user to learn two libraries at once. Many people are happy with Aeson and they might appreciate that they can apply that knowledge. That said, of course I think it would be nice to also show how to use Orville with `json-fleece`.

qxjit: No, I don't think it's ready now. We want to release orville before fleece will be ready

JSON.md
qxjit: I think for these tutorials it's better to show the already low-level `executeAndDecode` rather than `readRows` and `marshallResultFromSqlUsingRowIdExtractor`.

ysangkok: Fixed in 96db871f38075a137133f220f47f372ef4d436d4

qxjit (approved):

PR #266, opened 2023-02-28
Fix UTC offset sign

ysangkok:
Precedence confusion in the UTC time parser was causing nonsensical negation of only the hour, but not the minute component. Fixed in this PR.

An upcoming PR will add seconds support, if I can manage to do it, which I am not sure of, since ISO 8601 doesn't support the seconds, and that might mean doing too much time infrastructure ourselves.


qxjit (approved):
:heart:

PR #269, opened 2023-03-02
Compatibility with resource-pool 0.3/0.4

### LINKED ISSUE #259: Compat with resource-pool 0.3 and 0.4
Tests don't compile with these recentish release series. They removed MonadControlIO usage. We could maybe somehow convert the existing PropertyT to an IO which is what resource-pool takes? Unsure.

ysangkok:
Now that `Pool.withResource` returns a plain IO, apply the principle of least privilege and extract MonadTest code to outside of `HH.evalIO . Pool.withResource` blocks.

An advantage of this change is that it is now easier to see which parts of the code need `connection`, and which parts don't.

Fixes #259.

See also commercialhaskell/stackage#6852

### DIFF
--- a/GETTING-STARTED.md
+++ b/GETTING-STARTED.md
@@ -45,7 +45,6 @@ We need to tell Cabal how to find Orville:
 ```shell
 cat << EOF > cabal.project
 packages: .
-constraints: resource-pool < 0.3
 source-repository-package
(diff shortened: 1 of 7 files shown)


qxjit (approved):
:heart: Thank you!

PR #275, opened 2023-04-06
Organizes modules related to executing SQL operations under Execution

qxjit:


orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Insert.hs
```diff
@@ -6,7 +6,7 @@ Module    : Orville.PostgreSQL.Internal.Insert
 Copyright : Flipstone Technology Partners 2021
 License   : MIT
 -}
-module Orville.PostgreSQL.Internal.Insert
+module Orville.PostgreSQL.Execution.Insert
```
ysangkok, line 9: Now the comment at the top of the file is out of sync. Wouldn't it be best to just remove these comments? The copyright applies even if there isn't a notice in each file.

jbrechtel (approved):
:heart:

orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Insert.hs
qxjit: @ysangkok For now I'm just going to remove the `Module:` attribute from these comments. @telser and I will discuss further the implications of removing the other info from the comments. We may decide that we want the info to appear it haddock.

PR #277, opened 2023-04-17
Require time >=1.9.1

ysangkok:
The use of [nominalDiffTimeToSeconds](https://hackage.haskell.org/package/time-1.12.2/docs/Data-Time-Clock.html#v:nominalDiffTimeToSeconds)
means that the code won't compile on versions before time-1.9.1.


qxjit (approved):
:heart:

PR #278, opened 2023-04-17
Require attoparsec >=0.10

ysangkok:
The use of the `Data.Attoparsec.Bytestring` module means that the code will not compile before attoparsec 0.10. It is [not available in 0.9.1.2](https://hackage.haskell.org/package/attoparsec-0.9.1.2).


qxjit (approved):
:heart:

PR #280, opened 2023-05-02
Adds Haddocks to several modules.

telser:
Haddocks, primarily module headers, are added to AutoMigration, EntityTrace, ErrorDetailLevel, and Execution/*. This includes a module description, which is short, but shows the general pattern for doing so. I suspect that we will want to expand that pattern going forward.

This is a somewhat arbitrary place to cut this work, and more is to follow.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
@@ -1,10 +1,15 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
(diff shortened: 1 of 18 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
```diff
@@ -1,10 +1,15 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
telser, line 4: Oh right, we had some extensions that weren't used, so I cleaned them up.

qxjit (approved):

PR #286, opened 2023-05-09
Exports new clearTracesInState

ysangkok:
This makes it possible to clear the entity trace without exposing its internals.


qxjit (approved):
:heart:

PR #287, opened 2023-05-12
Make it possible to use SelectList without SqlMarshaller

ysangkok:
If a user wants to use SelectList without a marshaller, there is currently no
way to do that without copying the definition of `selectSelectList`.

This PR makes it possible to do that by, by removing the marshalling related
parts, which can simply be called independently.

I thought I might move it to Expr, since it doesn't actually execute anything,
but I left the function in the current module, since it depends on
SelectOptions, which is in Execution, and I thought it would be bad to have
Expr depend on Execution.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Select.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Select.hs
@@ -15,7 +15,6 @@ module Orville.PostgreSQL.Execution.Select
     selectToQueryExpr,
     selectTable,
     selectMarshalledColumns,
-    querySelectList,
     rawSelectQueryExpr,
   )
 where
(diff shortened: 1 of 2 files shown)


qxjit:
@ysangkok I think the issue isn't using `SelectList` without a `SqlMarshaller` (you would usually one or the other anyway, not both) but rather using `SelectOptions`. It seems what we're missing is a function that is the equivalent of `Expr.queryExpr` but for building a `QueryExpr` when you have a select options in hand. I think you've written that function, but it should probably live with `SelectOptions` rather than in the `Select` module, and maybe be named `selectOptionsQueryExpr` or something like that?

ysangkok:
@qxjit Applied your suggestion in 862359f

qxjit (approved):

PR #290, opened 2023-05-22
Add UnliftIO helpers, export Cursor type

ysangkok:
These helpers make it easier to use Cursors with UnliftIO, since the cursor can automatically get cleaned up in the case of exceptions.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
 
 {- |
 Copyright : Flipstone Technology Partners 2023
@@ -8,12 +9,15 @@ Stability : Stable
 @since 0.10.0.0
 -}
 module Orville.PostgreSQL.Execution.Cursor
-  ( withCursor,
+  ( Cursor,
+    withCursor,
     declareCursor,
     closeCursor,
     newCursorName,
     fetch,
     move,
+    acquireCursor,
+    acquireCursorUnlift,
     Expr.CursorDirection,
     Expr.next,
     Expr.prior,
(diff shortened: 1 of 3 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
```diff
@@ -194,3 +200,24 @@ newCursorName =
         "orville_cursor_%x_%08x"
         nowAsInteger
         (randomWord :: Word.Word32)
+
+acquireCursor ::
```
telser, line 204: Can we get haddock for these? Similar to `move` above?

orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
```diff
@@ -34,6 +38,8 @@ where
 
 import Control.Exception (bracket)
 import Control.Monad.IO.Class (MonadIO (liftIO))
+import Control.Monad.IO.Unlift (MonadUnliftIO (withRunInIO))
```
telser, line 41: Because this ends up so close to the `Control.Exception` import, I do wonder if we should switch to using the unliftio version of `bracket` (and others were necessary).

telser (changes requested):
Really just the haddock thing to change, everything else is commentary/questions.

orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
qxjit: @telser Using unliftio's bracket would mean requiring a `MonadUnliftIO` instance in order to use `withCursor` rather that using our own `liftBracket` like we do everywhere else.

qxjit:
@ysangkok On further reflection, I think we're not ready to start offering `resourcet` helpers directly from the orville package. These functions can easily by implemented externally by anyone trying to use `resourcet` and `orville` together, at least for now. Could we close this and open a separate PR that just exposes the `Cursor` type?

orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/Cursor.hs
ysangkok: Added in 262f59883f1dc88af6a75e0e3bd723f8284c0ff1

ysangkok:
Closing as requested! Copy these definitions to the use context if required.

PR #292, opened 2023-05-24
Fix incorrect ordering of order by and group by

exodrifter:
This fixes a malformed syntax error which occurs when you try to write a query which uses both a `GROUP BY` and `ORDER BY` clause. I've added tests verifying that the clauses can be used together.

I also noticed the group by tests were not added to the test suite, but they didn't work because the order of the expected result was incorrect. I fixed this by correcting the order, but since the order in SQL is technically undefined when a sort has not been specified, I don't think this is the best fix. I attempted changing the tests to use a `Set` instead, but that necessitated a change to `assertEqualSqlRows` too, so I decided not to do that to keep the PR limited in size.


qxjit (approved):

qxjit:
:heart:

PR #294, opened 2023-05-31
Don't clear transaction stack when clearing traces

ysangkok:
If we clear the tx stack, an eventual transaction callback like e.g. `commitTransaction` (also in `EntityTrace`) will throw when it finds that there is no stack.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
@@ -330,6 +330,17 @@ data TransactionStack trace
   | SavepointCreated O.Savepoint (DList.DList trace) (TransactionStack trace)
 
 {- |
+  INTERNAL: Clears all the traces in the stack.
+-}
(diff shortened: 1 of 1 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/EntityTrace.hs
```diff
@@ -636,7 +636,7 @@ clearTracesInState ::
   m ()
 clearTracesInState askTraceState = do
   traceState <- askTraceState
-  liftIO $ atomicModifyIORef'_ (_entityTraceStateRef traceState) (const emptyTraceData)
+  liftIO $ atomicModifyIORef'_ (_entityTraceStateRef traceState) (\x -> x {committedTracesDList = mempty})
```
qxjit, line 639: Should this go through the transaction stack and also clear the uncommitted traces as well? Or perhaps it should be named `clearCommittedTracesInState`?

ysangkok: Cleared uncommitted traces as well in commit f77b0a4f08412f164af3cb49449e4eeaf171a70c

PR #296, opened 2023-06-10
Add EntityOperations.findEntities

ysangkok:
A common operation is to find entities given a list of their IDs. This is a generalization of `findEntity`.

The returned list is a list, not a non-empty. This is analogous to how the returned value of `findEntity` is a Maybe.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Schema/PrimaryKey.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Schema/PrimaryKey.hs
@@ -169,10 +169,9 @@ primaryKeyEquals keyDef key =
   primary keys.
 -}
 primaryKeyIn :: PrimaryKey key -> NonEmpty key -> Expr.BooleanExpr
-primaryKeyIn keyDef keys =
-  ExtraNonEmpty.foldl1'
-    Expr.orExpr
-    (primaryKeyEquals keyDef <$> keys)
+primaryKeyIn keyDef =
+  ExtraNonEmpty.foldl1' Expr.orExpr
+    . fmap (primaryKeyEquals keyDef)
 
 {- |
   INTERNAL: builds the where condition for a single part of the key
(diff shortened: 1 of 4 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Schema/PrimaryKey.hs
```diff
@@ -161,6 +162,18 @@ primaryKeyEquals keyDef key =
...
+  'primaryKeyIn' builds a 'Expr.BooleanExpr' that will match rows where
+  the primary key is contained the given list. For single-field primary keys
+  this is equivalent to 'fieldIn', but 'primaryKeyIn' also handles composite
+  primary keys.
+-}
+primaryKeyIn :: PrimaryKey key -> NonEmpty key -> Expr.BooleanExpr
+primaryKeyIn keyDef keys =
+  ExtraNonEmpty.foldl1'
+    Expr.orExpr
+    (primaryKeyEquals keyDef <$> keys)
```
telser, line 175: I'd much rather us use prefix fmap here and eta reduce the keys away since that'll help the function be inlined.

ysangkok: Fixed in b6fd04da4c1e7892927331acc78643ed4491387a. I didn't know point-free style inlines more. Do you know of any references for that? I like to read more.

telser: From the [User's Guide section on inlining](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#inline-and-noinline-pragmas)
> Moreover, GHC will only inline the function if it is fully applied, where “fully applied” means applied to as many arguments as appear (syntactically) on the LHS of the function definition. For example:
>
>comp1 :: (b -> c) -> (a -> b) -> a -> c
>{-# INLINE comp1 #-}
>comp1 f g = \x -> f (g x)
>
>comp2 :: (b -> c) -> (a -> b) -> a -> c
>{-# INLINE comp2 #-}
>
> comp2 f g x = f (g x)
> The two functions comp1 and comp2 have the same semantics, but comp1 will be inlined when applied to two arguments, while comp2 requires three. This might make a big difference if you say
> 
> map (not `comp1` not) xs
> which will optimise better than the corresponding use of comp2.

PR #300, opened 2023-07-10
Update GHC versions that we test and support

onslaughtq:
Drops support for GHC 8.8 and drops support for all but the latest versions of GHC 8.10, 9.0, and 9.2. Adds support for GHC 9.4.5 and 9.6.2.

For GHC  9.4.5 and above, we are using `resource-pool` v 0.4.x. This led to a deprecation warning of `createPool` (which was marked deprecated in v0.3.x). The newer version now uses `newPool` with a `PoolConfig`

For GHC 9.6.x, `liftA2` is exported from Prelude, so importing it via `Control.Applicative` is redundant.


onslaughtq:
Since `resource-pool` v0.3.x doesn't have a way of specifying the number of stripes when creating a connection, should we update our dependencies to exclude this version?  

It only has `poolMaxResources` which is defined as:
"The maximum number of resources to keep open across all stripes. The smallest acceptable value is 1. Note: for each stripe the number of resources is divided by the number of capabilities and rounded up. Therefore the pool might end up creating up to N - 1 resources more in total than specified, where N is the number of capabilities."

telser:
> Since `resource-pool` v0.3.x doesn't have a way of specifying the number of stripes when creating a connection, should we update our dependencies to exclude this version?

This seems like exactly what we should do particularly with the resource usage you quoted.

orville-postgresql-libpq/src/Orville/PostgreSQL/PgCatalog/DatabaseDescription.hs
```diff
@@ -13,7 +15,10 @@ module Orville.PostgreSQL.PgCatalog.DatabaseDescription
   )
 where
 
+#if MIN_VERSION_base(4,18,0)
```
ysangkok, line 18: you can write `!MIN_VERSION_base`. `#else` not needed.

onslaughtq: Personally, I find it easier to comprehend without the negation

onslaughtq:
@telser I've updated this PR to restrict resource-pool as mentioned.

ysangkok (approved):

telser (approved):

PR #302, opened 2023-07-26
[SC-30704] Adds support for jsonb

onslaughtq:
Adds a jsonb DataType, Field, and SqlType. No default value was given for the `jsonbField`.

Tests were added for the jsonb field and SqlType. For the jsonb field test, I chose to limit the keys and values to alpha-numeric values, despite the fact that keys and values can be any valid string.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
@@ -76,6 +76,7 @@ module Orville.PostgreSQL.Marshall.FieldDefinition
   , utcTimestampField
   , localTimestampField
   , uuidField
+  , jsonbField
   , fieldOfType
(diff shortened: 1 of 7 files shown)


qxjit (approved):

orville-postgresql-libpq/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -417,6 +414,15 @@ fixedTextField name len = fieldOfType (SqlType.fixedText len) name
 textSearchVectorField :: String -> FieldDefinition NotNull T.Text
 textSearchVectorField = fieldOfType SqlType.textSearchVector
 
+{- |
+  Builds a 'FieldDefinition' that stores Haskell 'T.Text' values as the
+  PostgreSQL "JSOBN" type.
```
ysangkok, line 419: Typo here

PR #303, opened 2023-08-01
Reworking qualification, adding qualification for column names

telser:
The previous qualification function is replaced by 'qualifyTable' and 'qualifySequence'. These are more constrained in type as the previous allowed for incorrect qualifications.

Further added is 'qualifyColumn' to allow columns to be correctly qualified by both schema and table.

Also any _ prefix record fields in modules associated were changed to match the i_ prefix used elsewhere.

For sequences and tables, we have a higher level "Identifier" interface, which tracks the schema separately, and is likely the interface most will want to use. Because of that the functions 'qualifySequence' and 'qualifyTable' both take 'Maybe SchemaName'. This raises a couple of direction questions though:

- Do we want a similar interface for ColumnName?

  Sequences and tables both live directly on a schema, so there is a difference at least there. And on the other end is a 'FieldDefinition', which can _feel_ a bit like providing the identification of a column. But that is not exactly the same thing, notably 'FieldDefition' do not keep a reference to the parent like the the *Identifier types do, and providing that is likely not something we want.

- Should we simply change the higher interfaces to provide something like 'Either TableName (Qualified TableName)' and deal with that complexity?

  This has lots of obvious drawbacks of additional complexity where
today we simply create a 'Qualified TableName' or 'Qualified SequenceName' that lacks the qualification if the 'SchemaName' isn't provided. It definitely seems more directly usable to not have that complexity, but then the interface of 'Qualified Type' is a bit of a lie.

Given those questions, this takes the easy path forward and simply punts on answering any of them.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
@@ -1,17 +1,23 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 {- |
-Copyright : Flipstone Technology Partners 2016-2021
+Copyright : Flipstone Technology Partners 2016-2023
 License   : MIT
+Stability : Stable
 -}
 module Orville.PostgreSQL.Expr.Internal.Name.Qualified
   ( Qualified
-  , qualified
+  , qualifyTable
+  , qualifySequence
+  , qualifyColumn
   )
 where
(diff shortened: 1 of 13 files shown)


qxjit (approved):
:heart:

qxjit:
It's attractive to have a `ColumnIdentifier`, but trick to sort out how it works with `FieldDefinition`, especially since a single `FieldDefinition` can be re-used across multiple tables. Having an table or schema qualification inside it seems inappropriate for that situation.

PR #305, opened 2023-08-01
Require BooleanExpr for updateFields

ysangkok:
Using `UPDATE` without a where clause is an edge case, which shouldn't hurt the usability of people executing `UPDATE` with a where clause. EntityOperations should be optimized for the common case. By requiring the BooleanExpr instead of having it be Maybe, usability is improved for users that always have where clauses, since they won't need to supply the Just constructor or define a helper to do that.

The small minority of queries that actually need to be executed without a where clause can use custom operations. If a trivially true clause isn't a problem `literalBooleanExpr` can be used.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/EntityOperations.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Execution/EntityOperations.hs
@@ -163,11 +163,11 @@ updateFields ::
   Monad.MonadOrville m =>
   Schema.TableDefinition (Schema.HasKey key) writeEntity readEntity ->
   NonEmpty Expr.SetClause ->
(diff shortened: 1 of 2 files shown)


qxjit:
I'm in favor of this is spirit, but I don't really like forcing users into passing a dummy boolean expr when they want to run a sql UPDATE with no where clause whatsoever.

ysangkok:
I think we reached consensus to change the Maybe to be something like `data HasWhereCondition = WithWhere | WithoutWhere`, such that `deleteWhere ... Nothing`  doesn't sound so much like it would delete nothing while actually deleting everything.

Another change I will submit separately is have row count returning functions use a suffix similar to the `AndReturnEntity` variants we already have.

PR #307, opened 2023-08-10
Short EntityOperations functions always produce ()

ysangkok:
It's very common to not need the row count returned by these functions, since one assumes that the database operation failed if not all the given rows were inserted. Having these operations return unit avoids having to do `_ <-` or using `Monad.void`.


qxjit (approved):

telser:
Maybe we should change the `()` ones to have a RETURNING NULL clause so that they truly don't get anything back over the wire?

PR #313, opened 2023-08-30
Adding yet more Haddocks.

telser:
Also fixes a couple of docs with mistakes that came in with 2b4d18de18651ab7f9d80a313be29d6eeb15d191.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
@@ -24,7 +24,7 @@ Type to represent a SQL identifier. 'Identifier' values constructed via the
 
 There is an low level escape hatch included here, by means of the instance of
 'RawSql.SqlExpression'. This is intended to be used when some functionality is
-required but not already included. The exension mechanism provided does require
+required but not already included. The extension mechanism provided does require
 care in use as no guarantees are provided for correctness in usage.
 
 For example, if one wanted to write a raw (unescaped) identifier by hand and
@@ -59,13 +59,16 @@ identifierFromBytes :: B8.ByteString -> Identifier
 identifierFromBytes =
   Identifier . RawSql.identifier
 
-{- |
-
+{- | This class aids in giving additional polymorphism and so that many different identifiers can be
+created without being forced to only use the `Identifier` type.
 
 @since 0.10.0.0
 -}
 class IdentifierExpression name where
+  -- | @since 0.10.0.0
   toIdentifier :: name -> Identifier
+
+  -- | @since 0.10.0.0
   fromIdentifier :: Identifier -> name
 
 {- |
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
@@ -1,8 +1,11 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 {- |
-Copyright : Flipstone Technology Partners 2016-2021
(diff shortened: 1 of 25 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
```diff
@@ -38,18 +41,37 @@ newtype Identifier
...
+@since 0.10.0.0
+-}
 class IdentifierExpression name where
   toIdentifier :: name -> Identifier
   fromIdentifier :: Identifier -> name
 
+{- |
+
+@since 0.10.0.0
+-}
```
AugmenTab, line 74: Were these meant to have other copy besides the @since?

orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/OrderBy.hs
```diff
@@ -22,65 +25,164 @@ module Orville.PostgreSQL.Expr.OrderBy
...
 
 import Orville.PostgreSQL.Expr.Name (ColumnName)
 import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
 
+{- |
+Type to represent a SQL order by clause.
+
+There is an low level escape hatch included here, by means of the instance of
+'RawSql.SqlExpression'. This is intended to be used when some functionality is
+required but not already included. The exension mechanism provided does require
```
AugmenTab, line 37: Is "exension" here meant to be "extension?"

orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/OrderBy.hs
```diff
@@ -22,65 +25,164 @@ module Orville.PostgreSQL.Expr.OrderBy
...
 orderByClause :: OrderByExpr -> OrderByClause
 orderByClause expr = OrderByClause (RawSql.fromString "ORDER BY " <> RawSql.toRawSql expr)
 
+{- |
+Type to represent a SQL order by expression (the part that follows the @ORDER
+BY@ in sql).
+
+There is an low level escape hatch included here, by means of the instance of
+'RawSql.SqlExpression'. This is intended to be used when some functionality is
+required but not already included. The exension mechanism provided does require
```
AugmenTab, line 60: Same as above.

orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/OrderBy.hs
```diff
@@ -22,65 +25,164 @@ module Orville.PostgreSQL.Expr.OrderBy
...
 orderByColumnName :: ColumnName -> OrderByDirection -> OrderByExpr
 orderByColumnName =
   orderByExpr . RawSql.toRawSql
 
+{- |
+Type to represent a SQL order by direction expression.
+
+There is an low level escape hatch included here, by means of the instance of
+'RawSql.SqlExpression'. This is intended to be used when some functionality is
+required but not already included. The exension mechanism provided does require
```
AugmenTab, line 126: Same as above.

telser: Thanks for catching these!

orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Internal/Name/Identifier.hs
telser: I only added the `@since` and didn't put anything else on the class or the methods because I couldn't come up with anything that wasn't a restating of the method names. But thinking about it more saying something about being more polymorphic could be good and we should at least put the `@since` on the class methods too.

telser:
@AugmenTab I think all of your comments should have been addressed now

AugmenTab (approved):

PR #315, opened 2023-09-15
Documents `planList` behavior for input and result lists

qxjit:
This add some details to the docs for `planList` to explain that the
input and result lists are be correlated.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
@@ -366,9 +366,17 @@ planMany =
   PlanMany
 
 {- |
-  'planList' lifts a plan so both its param and result become lists.
-  This saves you from having to fmap in 'Many.elems' when all you want back
-  from a 'Many' is the list of results inside it.
+  'planList' lifts a plan so both its param and result become lists. This saves
+  you from having to fmap in 'Many.elems' when all you want back from a 'Many'
+  is the list of results inside it.
+
+  There will always be the same number of elements in the @[result]@ list as
+  there are in the @[param]@ list, even if there are duplicate values in the
+  input parameters. This may be counter-intuitive in the trivial case where
+  plan that queries a single table is passed to 'planList' but cannot be
+  avoided due to more complicated situations where the original plan executes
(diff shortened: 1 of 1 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
```diff
@@ -366,9 +366,17 @@ planMany =
...
-  This saves you from having to fmap in 'Many.elems' when all you want back
-  from a 'Many' is the list of results inside it.
+  'planList' lifts a plan so both its param and result become lists. This saves
+  you from having to fmap in 'Many.elems' when all you want back from a 'Many'
+  is the list of results inside it.
+
+  There will always be the same number of elements in the @[result]@ list as
+  there are in the @[param]@ list, even if there are duplicate values in the
+  input parameters. This may be counter-intuitive in the trivial case where
+  plan that queries a single table is passed to 'planList' but cannot be
```
AugmenTab, line 376: Did you mean "where a plan that queries" here?

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
```diff
@@ -366,9 +366,17 @@ planMany =
...
+  avoided due to more complicated situations where the original plan executes
+  queries against multiple tables the query results must be correlated based on
```
AugmenTab, line 378: Was "multiple tables" meant to be the end of the sentence here, with "the query results" starting a new sentence? I'm not sure what this is trying to convey.

jbrechtel (approved):

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
qxjit: That wasn't the intention, but I can see how it might be easier to understand if I separate and start a new sentence with "In  this situation the query results..."

orville-postgresql-libpq/src/Orville/PostgreSQL/Plan.hs
qxjit: Yes, thank you. :heart:

PR #322, opened 2023-09-21
Adds basic support for concurrent index creation

qxjit:
This adds the option for a user to specify an `IndexDefinition` should
be created asynchronously during migrations.

To do this the migration lock needed to be changed to be acquired at the
session level rather than the transaction level, which means we need to
manually release the lock. I added tests to ensure that this lock
release happens as expected. I also exported the `withTransactionLock`
function both to test it, and to make it available to users in case they
want to use `generateMigrationPlan` and `executeMigrationPlan`
separately.

### DIFF
--- a/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Index.hs
+++ b/orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Index.hs
@@ -11,6 +11,10 @@ module Orville.PostgreSQL.Expr.Index
   ( CreateIndexExpr
   , createIndexExpr
   , IndexUniqueness (UniqueIndex, NonUniqueIndex)
+  , IndexBodyExpr
+  , indexBodyColumns
+  , ConcurrentlyExpr
+  , concurrently
   , DropIndexExpr
   , dropIndexExpr
   , createNamedIndexExpr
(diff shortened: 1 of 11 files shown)


orville-postgresql-libpq/src/Orville/PostgreSQL/Expr/Index.hs
```diff
@@ -45,48 +49,103 @@ newtype CreateIndexExpr
...
+@since 0.10.0.0
+-}
+concurrently :: ConcurrentlyExpr
+concurrently =
+  RawSql.unsafeSqlExpression "CONCURRENTLY"
+
+{- |
+Type to represent if the body of an index definition (i.e. @<body>@ in
+after @CREATE some_index ON some_table <body>@).
+
```
telser, line 125: I've been adding the same verbiage about the `SqlExpression` escape hatch to these newtypes. Would it be good to include here for cases we don't currently support? Obviously one _could_ use the higher up `CreateIndexExpr`, but something for the WITH/INCLUDE/TABLESPACE/WHERE options might be useful to someone at this level too.

qxjit: @telser I updated the `SqlExpression` docs and then add references to them on the newly added types.

telser (approved):

PR #329, opened 2023-10-05
Add some missing exports

deathgrindfreak:
We're upgrading to the latest version of Orville in SPUR and I believe these are the only exports that we had to have.


qxjit:
@deathgrindfreak Would it be reasonable for you to access these by importing `Orville.PostgreSQL.Schema` instead?

deathgrindfreak:
@qxjit Sorry, was in a bit of a hurry to get on with vacation.  That'll work just fine 🤘

PR #337, opened 2023-10-27
Adds `ConnectionOptions` for `createConnectionPool` parameters

qxjit:
Providing a `ConnectionOptions` record should be a better experience for
anyone trying to read the docs or look at example code. This also adds a
`StripeOption` type that allows users to easily get one stripe allocated
per thread capability, which is what is recommended by `resource-pool`.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
@@ -9,23 +9,30 @@ Stability : Stable
 @since 1.0.0.0
 -}
 module Orville.PostgreSQL.Raw.Connection
-  ( -- * Orville definitions
-    Connection
-  , createConnectionPool
+  ( ConnectionOptions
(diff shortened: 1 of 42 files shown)


onslaughtq (approved):

orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
```diff
@@ -55,43 +62,113 @@ data NoticeReporting
...
+-}
+determineStripeCount :: StripeOption -> IO Int
+determineStripeCount stripeOption =
+  case stripeOption of
+    OneStripePerCapability -> getNumCapabilities
+    StripeCount n -> pure n
+
+{- |
+  Allocates a connection from the pool and performs an action with it. This
+  function will block if the maximum number of connections is reached.
```
telser, line 159: Missing the since annotation here.

telser (approved):

PR #341, opened 2024-02-12
Add 'toRawBytes' in SqlValue

ysangkok:
This could be useful if e.g. someone is using the 'bytea' type. Why force people to use Text/String if the API supports ByteStrings just fine?


ysangkok:
Closing because this is already offered by `toPgValue`

PR #344, opened 2024-02-22
Adds a basic 'OnConflictExpr'

telser:
This adds very rudementary support for ON CONFLICT in the 'InsertExpr'. Adding only a type and a helper for 'ON CONFLICT DO NOTHING'.

The placement of the type alongside the insert related is somewhat arbitrary compared to being a separate module. It was placed within the insert because it would appear that postgresql, as of version 16, only supports it on insert statements so it is not really reusable elsewhere.

This also changes the 'insertExpr' function to optionally take the new 'OnConflictExpr'. As well as the 'mkInsertExpr' at the 'TableDefinition' level. However, this is not filtered to the highest levels of the api. The lower levels are changed because the ordering of the 'ON CONFLICT', must come before the 'RETURNING' part of an insert and without the change the only way to effectively use the change would have been to not use the existing expr level functionality at all. The highest level is unchanged so we think through how to evolve the api, particularly should we want good support for 'ON CONFLICT DO SET ...'. This set of low level/high level feels like a decent tradeoff but revisiting the decisions is without question, appropriate too.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Insert.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Insert.hs
@@ -16,6 +16,8 @@ module Orville.PostgreSQL.Expr.Insert
   , insertSqlValues
   , RowValues
   , rowValues
+  , OnConflictExpr
+  , onConflictDoNothing
   )
 where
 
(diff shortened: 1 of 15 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/Insert.hs
```diff
@@ -168,3 +172,26 @@ rowValues values =
...
+
+'OnConflict' provides a 'RawSql.SqlExpression' instance. See 'RawSql.unsafeSqlExpression' for how to
+construct a value with your own custom SQL.
+
+@since 1.1.0.0
+-}
+newtype OnConflictExpr
+  = OnConflictExpr RawSql.RawSql
+  deriving
+    ( -- | @since 1.0.1.0
```
qxjit, line 187: This `@since` doesn't match the one on the type.

telser: That you this should be updated now!

qxjit (approved):

PR #346, opened 2024-02-27
Adds initial aliasing support.

telser:
The highlights:

- Moving to use 'Qualified ColumnName' in many/most places. This allows for the aliasing to be used naturally in those places.

- Add an alias type and function to create a 'Qualified ColumnName' with it. This signals that 'Alias' is in fact different than 'ColumnName' or 'TableName', and reusing them for aliasing is not always appropriate.

- Adding a 'MarshallAlias' constructor to the 'SqlMarshaller' so that an alias can be attached to all of the natural fields in a marshaller in a single function call from users. Synthetic fields could be refactored with this facility, but that was left out here to keep the already big changeset smaller.

- Many of the fieldDefinition helpers now have a version to support generating various subexpressions with an alias along with the fieldDefinition. This is a choice to allow the simple case to not deal with the complexity of the alias, but make it easier for when aliases are needed.

- A small number of helpers are added to make generation of bits of SQL, such as a 'TableReferenceList' that includes the 'AS alias' piece.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
@@ -426,10 +426,10 @@ fieldColumnReferenceWithAlias mbAlias =
 
 @since 1.0.0.0
 -}
-fieldColumnDefinition :: Maybe Expr.Alias -> FieldDefinition nullability a -> Expr.ColumnDefinition
-fieldColumnDefinition mbAlias fieldDef =
+fieldColumnDefinition :: FieldDefinition nullability a -> Expr.ColumnDefinition
+fieldColumnDefinition fieldDef =
   Expr.columnDefinition
-    (fieldColumnName mbAlias fieldDef)
+    (fieldNameToColumnName $ fieldName fieldDef)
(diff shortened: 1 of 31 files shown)


orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -891,14 +935,31 @@ fieldNotEquals =
...
 
+{- |
+  Checks that the value in an aliased field does not equal a particular value.
+
+@since 1.0.0.0
+-}
+fieldNotEqualsWithAlias :: Maybe Expr.Alias -> FieldDefinition nullability a -> a -> Expr.BooleanExpr
+fieldNotEqualsWithAlias mbAlias =
+  whereColumnComparisonWithAlias mbAlias Expr.notEquals
+
```
qxjit, line 946: I think we should revisit adding `WithAlias` versions of all these functions. I don't have an immediate suggestion, but having 2 versions of each function seems like a pretty inconvenient API for users

qxjit (changes requested):

telser:
Closing this in favor of #352.

PR #347, opened 2024-02-28
Adds initial support for triggers and functions.

qxjit:
This adds `Expr` functions for building SQL expressions related to
triggers and functions along with schema level `FunctionDefinition`
and `TriggerDefinition` abstractions.

`FunctionDefinition`s are meant be included directly in `SchemaItem`
lists for auto migration purposes to create the functions that triggers
will call. For the purposes of this change we provide only `mkTriggerFunction`
for construction a new `FunctionDefinition` for a trigger function.

`TriggerDefinition`s are added to a `TableDefinition` via the new
`addTableTriggers` function, similar to indexes and constraints.
For the time being migration of trigger's is done by name only --
existing triggers are never altered. This is due both to the complexity
involved in interpreting the `pg_trigger` catalog entries to determine
whether the trigger needs to be migration and also the fact that
`CREATE OR REPLACE TRIGGER` is not available in PostgreSQL 14 and
would be required for changing most of the properties on triggers.

Co-Authored-By: Paul <paul@flipstone.com>

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -528,8 +528,8 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
                   AddFunctions
                   (Orville.mkCreateFunctionExpr Nothing functionDef)
               ]
-            Just proc ->
-              if PgCatalog.pgProcSource proc == T.pack (Orville.functionSource functionDef)
+            Just pgProc ->
+              if PgCatalog.pgProcSource pgProc == T.pack (Orville.functionSource functionDef)
                 then []
                 else
                   [ mkMigrationStepWithType
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -526,7 +526,7 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
             Nothing ->
               [ mkMigrationStepWithType
                   AddFunctions
-                  (Orville.mkCreateFunctionExpr Nothing functionDef)
+                  (Orville.mkCreateFunctionExpr functionDef Nothing)
               ]
             Just pgProc ->
               if PgCatalog.pgProcSource pgProc == T.pack (Orville.functionSource functionDef)
@@ -534,7 +534,7 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
                 else
                   [ mkMigrationStepWithType
                       AddFunctions
-                      (Orville.mkCreateFunctionExpr (Just Expr.orReplace) functionDef)
+                      (Orville.mkCreateFunctionExpr funcitonDef (Just Expr.orReplace))
                   ]
     SchemaDropFunction functionId ->
       Right $
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -534,7 +534,7 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
                 else
                   [ mkMigrationStepWithType
                       AddFunctions
-                      (Orville.mkCreateFunctionExpr funcitonDef (Just Expr.orReplace))
+                      (Orville.mkCreateFunctionExpr functionDef (Just Expr.orReplace))
                   ]
     SchemaDropFunction functionId ->
       Right $
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -3,7 +3,7 @@
 {-# LANGUAGE RankNTypes #-}
 
 {- |
-Copyright : Flipstone Technology Partners 2023
+Copyright : Flipstone Technology Partners 2023-2024
 License   : MIT
 Stability : Stable
 
@@ -42,6 +42,7 @@ import qualified Data.Map.Strict as Map
 import qualified Data.Maybe as Maybe
 import qualified Data.Set as Set
 import qualified Data.String as String
+import qualified Data.Text as T
 import qualified Data.Text.Encoding as Enc
 import qualified Database.PostgreSQL.LibPQ as LibPQ
 
@@ -88,6 +89,19 @@ data SchemaItem where
   SchemaDropSequence ::
     Orville.SequenceIdentifier ->
     SchemaItem
+  -- |
+  --    Constructs a 'SchemaItem' from a 'Orville.FunctionDefinition'.
+  -- @since 1.1.0.0
+  SchemaFunction ::
+    Orville.FunctionDefinition ->
+    SchemaItem
+  -- |
+  --    Constructs a 'SchemaItem' that will drop the specified table if it is
+  --    found in the database.
+  -- @since 1.1.0.0
+  SchemaDropFunction ::
+    Orville.FunctionIdentifier ->
+    SchemaItem
 
 {- |
   Returns a one-line string describing the 'SchemaItem', suitable for a human
@@ -109,6 +123,10 @@ schemaItemSummary item =
       "Sequence " <> Orville.sequenceIdToString (Orville.sequenceIdentifier sequenceDef)
     SchemaDropSequence sequenceId ->
       "Drop sequence " <> Orville.sequenceIdToString sequenceId
+    SchemaFunction functionDef ->
+      "Function " <> Orville.functionIdToString (Orville.functionIdentifier functionDef)
+    SchemaDropFunction functionId ->
+      "Drop function " <> Orville.functionIdToString functionId
 
 {- |
 A 'MigrationPlan' contains an ordered list of migration steps. Each one is a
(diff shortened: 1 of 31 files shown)


orville-postgresql/src/Orville/PostgreSQL/PgCatalog/DatabaseDescription.hs
```diff
@@ -65,6 +71,18 @@ lookupRelation ::
 lookupRelation key =
   Map.lookup key . databaseRelations
 
+{- |
+  Lookup a procedure by its qualified name in the @pg_catalog@ schema.
+
+@since 1.0.0.0
```
onslaughtq, line 77: I think this should be since 1.1.0.0

qxjit: Fixed, thanks!

orville-postgresql/src/Orville/PostgreSQL/Expr/Function.hs
```diff
@@ -0,0 +1,257 @@
...
+  CreateFunctionExpr
+createFunction maybeOrReplace name functionReturns functionLanguage definition =
+  CreateFunctionExpr $
+    RawSql.intercalate
+      RawSql.space
+      [ RawSql.fromString "CREATE"
+      , maybe mempty RawSql.toRawSql maybeOrReplace
+      , RawSql.fromString "FUNCTION"
+      , RawSql.toRawSql name
+      , RawSql.fromString "()" -- currently we don't support specifying arguments
```
telser, line 109: Maybe this should be in the haddocks?

orville-postgresql/src/Orville/PostgreSQL/Expr/Trigger.hs
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+{- |
+Constructs a SQL @CREATE TRIGGER@ statement from the trigger name and its defining
+attributes.
+
+@since 1.1.0.0
+-}
+createTrigger ::
+  Maybe OrReplace ->
```
telser, line 104: Should we state in the haddock that 'Nothing' is the only valid value here for versions prior to 14?

orville-postgresql/src/Orville/PostgreSQL/Expr/Trigger.hs
```diff
@@ -0,0 +1,279 @@
...
+      , RawSql.fromString "TRIGGER"
+      , RawSql.toRawSql name
+      , RawSql.toRawSql timing
+      , RawSql.intercalate (RawSql.fromString " OR ") events
+      , RawSql.fromString "ON"
+      , RawSql.toRawSql tableName
+      , RawSql.toRawSql fireScope
+      , RawSql.fromString "EXECUTE FUNCTION"
+      , RawSql.toRawSql functionName
+      , RawSql.fromString "()" -- we don't currently support objects
```
telser, line 127: Might be worth mentioning this in the docs too.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -482,6 +516,40 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
...
+            functionIdToPgCatalogNames currentNamespace (Orville.functionIdentifier functionDef)
+        in
+          case PgCatalog.lookupProcedure (schemaName, procName) dbDesc of
+            Nothing ->
+              [ mkMigrationStepWithType
+                  AddFunctions
+                  (Orville.mkCreateFunctionExpr Nothing functionDef)
+              ]
+            Just proc ->
+              if PgCatalog.pgProcSource proc == T.pack (Orville.functionSource functionDef)
```
telser, line 532: Just a note: proc is a keyword with some extensions turned on.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -609,13 +683,36 @@ mkAlterTableSteps currentNamespace relationDesc tableDef =
...
+        (PgCatalog.relationTriggers relationDesc)
+
     tableName =
       Orville.tableName tableDef
   in
     mkAlterColumnSteps tableName (addAlterColumnActions <> dropColumnActions)
       <> mkConstraintSteps tableName (addConstraintActions <> dropConstraintActions)
       <> addIndexSteps
       <> dropIndexSteps
+      <> addTriggerSteps
```
telser, line 714: Would anything ever go in the middle of the add/drop of the trigger steps here? I'm wondering if we could or should do the grouping like what happens for the constraints and columns above for triggers and indexes.

telser (commented):
Looks pretty good. Had some minor things about docs. Really like how we're able to get this and still support the automigrations.

orville-postgresql/src/Orville/PostgreSQL/PgCatalog/PgProc.hs
```diff
@@ -0,0 +1,103 @@
...
+  The Haskell representation of data read from the @pg_catalog.pg_proc@ table.
+  Rows in this table reperesent functions, procedures, aggregate functions, and
+  window functions (collectively also known as routines), such as those created
+  via @CREATE FUNCTION@ or @CREATE PROCEDUCE.
+
+@since 1.1.0.0
+-}
+data PgProc = PgProc
+  { pgProcOid :: LibPQ.Oid
+  -- ^ The PostgreSQL @oid@ for the relation.
```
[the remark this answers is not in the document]
qxjit: I thought that was pretty surperfluous when all the fields were introduced at the same time as the record. I don't think we need to `@since`  every field of every record do we?

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
qxjit: Remember the the order things appear in this this is not the order that they're executed in -- that order is determined by the `StepType` that the migration step was created with. So the grouping here doesn't really matter at all.

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionDefinition.hs
```diff
@@ -0,0 +1,122 @@
...
+  String ->
+  FunctionDefinition ->
+  FunctionDefinition
+setFunctionSchema schemaName functionDef =
+  functionDef
+    { i_functionIdentifier = setFunctionIdSchema schemaName (i_functionIdentifier functionDef)
+    }
+
+{- |
+  Retrieves the 'FunctionIdentifier' for this sequence, which is set by the
```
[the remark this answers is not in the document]
qxjit: This was already addressed in [e1d53d2](https://github.com/flipstone/orville/pull/347/commits/e1d53d2c75f82c7c3f9a51a2148d8ffe6d1f1ae6)

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionDefinition.hs
```diff
@@ -0,0 +1,122 @@
...
+mkTriggerFunction name language source =
+  FunctionDefinition
+    { i_functionIdentifier = unqualifiedNameToFunctionId name
+    , i_functionReturnType = Expr.returnTypeTrigger
+    , i_functionLanguage = language
+    , i_functionSource = source
+    }
+
+{- |
+  Builds a 'Expr.CreateFunctionExpr' that will create a SQL sequence matching
```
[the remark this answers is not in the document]
qxjit: Already fixed [e1d53d2](https://github.com/flipstone/orville/pull/347/commits/e1d53d2c75f82c7c3f9a51a2148d8ffe6d1f1ae6)

orville-postgresql/src/Orville/PostgreSQL/Expr/Trigger.hs
qxjit: Addressed in [b0f56fd](https://github.com/flipstone/orville/pull/347/commits/b0f56fd0282cf789d6d355ec7e308d7b51fc12a8)

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionDefinition.hs
```diff
@@ -0,0 +1,122 @@
...
+  ) where
+
+import qualified Orville.PostgreSQL.Expr as Expr
+import Orville.PostgreSQL.Schema.FunctionIdentifier (FunctionIdentifier, functionIdQualifiedName, setFunctionIdSchema, unqualifiedNameToFunctionId)
+
+{- |
+  Contains the definition of a PostgreSQL function for Orville to use when creating
+  the function. Currently only functionality for creating trigger functions for use
+  with 'Orville.PostgreSQL.Schema.TriggerDefinition' is supported. You can create a
+  'FunctionDefintion' with for a trigger via 'mkTriggerFunction'.
```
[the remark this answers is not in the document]
qxjit: Addressed in [8e012e3](https://github.com/flipstone/orville/pull/347/commits/8e012e348b1364a8bd11f693ddc282bb579ebe0f)

orville-postgresql/src/Orville/PostgreSQL/Expr/Trigger.hs
qxjit: Addressed in [8e012e3](https://github.com/flipstone/orville/pull/347/commits/8e012e348b1364a8bd11f693ddc282bb579ebe0f)

onslaughtq:
I don't know if this worth updating, but one observation is that we weren't consistent in naming some functions, i.e `indexCreateExpr` vs `mkCreateSequenceExpr`.  It seems now that we added corresponding functions for triggers and functions, with `mk` at the front, we should rename `indexCreateExpr` to `mkCreateIndexExpr`.

Another observation is the placement of the `Maybe Expr.OrReplace` argument is second in `mkCreateTriggerExpr`, but first in `mkCreateFunctionExpr`.

PR #349, opened 2024-03-24
Adds basic support for PostgreSQL extensions and particularly pg_trgm

telser:
At the highest level this does four things:
- Adds support for extensions, including automigration
- Adds a 'IfNotExistsExpr' for use in some extension support.
- Adds support for a subset of functionality provided by the pg_trgm extension.
- Allows ordering to be done on a synthetic field.

Extension support is added at the expression level, as well as to the 'PgCatalog' tracking infrastructure to support 'AutoMigration' for loading/unloading extensions.

Support for cascading the load of extensions supported and used during automigration so that users can be sure to list the extension they want and any dependencies will be loaded as well.

To be used in conjection with the extension support is some functionality from pg_trgm. The ability to add both GIN and GIST indexes are included, as well as overriding the GIST index parameter, 'siglen'. Further, the similarity, word_similarity, and strict_word_similarity functions are exposed to compare value expressions. Higher level synthetic fields that compare a column with a value are also included. This hopefully covers a broad range of usage such as LIKE queries, column similarity, and the flexibility to use the function value in other ways.

Support for ordering by a synthetic field is added precisely be able sort by similarity.

pg_trgm has a few things **not** included here:

  - The notion of a similarity threshold for each of the three similarity functions. Including the ability to show and set the values.

  - Operators comparing a pair of values that each, implicitly use a corresponding threshold resulting in a boolean.

  - Operators that compute "distance", defined as 1 - similarity, for each of the three similarities.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -65,6 +65,7 @@ import qualified Orville.PostgreSQL.Schema as Schema
 data SchemaItem where
   -- |
   --    Constructs a 'SchemaItem' from a 'Orville.TableDefinition'.
+  --
   -- @since 1.0.0.0
   SchemaTable ::
     Orville.TableDefinition key writeEntity readEntity ->
@@ -72,12 +73,14 @@ data SchemaItem where
   -- |
   --    Constructs a 'SchemaItem' that will drop the specified table if it is
   --    found in the database.
+  --
   -- @since 1.0.0.0
   SchemaDropTable ::
     Orville.TableIdentifier ->
     SchemaItem
   -- |
   --    Constructs a 'SchemaItem' from a 'Orville.SequenceDefinition'.
+  --
   -- @since 1.0.0.0
   SchemaSequence ::
     Orville.SequenceDefinition ->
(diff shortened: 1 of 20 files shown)


orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -65,43 +65,63 @@ import qualified Orville.PostgreSQL.Schema as Schema
 data SchemaItem where
   -- |
   --    Constructs a 'SchemaItem' from a 'Orville.TableDefinition'.
+  --
```
telser, line 68: I forgot to mention this in the commit message, this is just a slight doc fix. Haddock doesn't pick up the since, without the empty line between. Which is.. a bit frustrating. But this can be seen on hackage for the release right now.

orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -1125,9 +1125,11 @@ whereColumnComparison columnComparison fieldDef a =
     (fieldColumnReference fieldDef)
     (fieldValueToExpression fieldDef a)
 
-{-- |
+{- |
```
telser, line 1128: The other documentation fix I found on hackage and forgot to mention in the commit message. This ends up not being haddock with the `--` instead of the `-`.

qxjit (approved):

PR #350, opened 2024-04-02
Adds the coalesce function

jlavelle:


ysangkok:
COALESCE [isn't actually a function](https://dba.stackexchange.com/a/335767/10479). It would be better if we used another term.

ysangkok:
I think Orville quotes function names when the SQL is generated. Since quoting won't work (see dba stackexchange link), that means the behaviour in this PR also wouldn't work. Or am I wrong about that somehow?

jlavelle:
You are absolutely right, thanks for catching this. That's what I get for not writing tests!

What about something like `coalesceFunction = RawSql.unsafeFromRawSql $ RawSql.fromString "coalesce"`? That side-steps the quoting issue but still treats `coalesce` as a function.

telser:
Since this isn't _really_ a function, I think probably need to model separately. PostgreSQL seems to call this and some others "conditional expressions" at https://www.postgresql.org/docs/current/functions-conditional.html which makes me lean towards that.

PR #351, opened 2024-04-02
Adds `orderByValueExpression`

jlavelle:


orville-postgresql/src/Orville/PostgreSQL/Expr/OrderBy.hs
```diff
@@ -100,6 +102,16 @@ orderByColumnName :: ColumnName -> OrderByDirection -> OrderByExpr
 orderByColumnName =
   curry (orderByColumnsExpr . pure)
 
+{-- |
+  Orders a query by the given 'ValueExpression' in the given order direction. The caller must
+  ensure that the 'ValueExpression' is valid as a sort expression.
+
+@since 1.1.0.0
+-}
+orderByValueExpression :: ValueExpression -> OrderByDirection -> OrderByExpr
```
telser, line 111: I think we probably shouldn't do something quite this general.

From the PostgreSQL [docs](https://www.postgresql.org/docs/current/sql-select.html#SQL-ORDERBY):
> A limitation of this feature is that an ORDER BY clause applying to the result of a UNION, INTERSECT, or EXCEPT clause can only specify an output column name or number, not an expression.

jlavelle: I'm struggling to imagine a less general/more safe solution that lets me do what I want (generate some SQL like `order by coalesce(x, 0)`). If we had support for `UNION`, `INTERSECT`, etc. maybe we'd have a separate `RestrictedOrderByExpr` or something.

telser: If I'm reading the docs correctly and not missing something from somewhere else, then it sounds like we should be able to order by the following two things: functions and case expressions, but not query combining expressions*. Then we could have targeted functions like `orderByFunction` and `orderByCaseExpression`.

* Perhaps `QueryCombiningExpr` is what we could name our model of the `UNION`, `INTERSECT` and  `EXCEPT` per the description here https://www.postgresql.org/docs/16/queries-union.html.

telser: I somehow managed to invert the requirements in my thinking above. We probably should get this in, maybe with the amendment to have the specialized versions call this.

jlavelle:
Forgot about this one. I rebased and rewrote the other order by functions in terms of `orderByValueExpression` as @telser suggested.

qxjit (approved):

telser (approved):

PR #352, opened 2024-04-06
Adds initial aliasing support- typeclassy

telser:
This is an updated/reworked version of #346. the primary difference between that and this is the creation of the typeclass using that to support comparisons on synthetic and aliased fields.

The highlights:

- Moving to use 'Qualified ColumnName' in many/most places. This allows for the aliasing to be used naturally in those places.

- Add an alias type and function to create a 'Qualified ColumnName' with it. This signals that 'Alias' is in fact different than 'ColumnName' or 'TableName', and reusing them for aliasing is not always appropriate.

- Adding a 'MarshallAlias' constructor to the 'SqlMarshaller' so that an alias can be attached to all of the natural fields in a marshaller in a single function call from users. Synthetic fields could be refactored with this facility, but that was left out here to keep the already big changeset smaller.

- Adds 'AliasedFieldDefinition' for holding a field that should be referenced with an alias in SQL.

- Adds a 'SqlComparable' typeclass to allow for comparison in a polymorphic way. Instances are provided for 'FieldDefinition', 'SyntheticField' and the new 'AliasedFieldDefinition'. Comparison functions are exposed working with the class. The field definition functions and operators for comparison are now type restricted versions of the classy functions. In this way the "simple" case of FieldDefinitions has the most specific functions, but 'AliasedFieldDefinition' and 'SyntheticField' can both be compared against.

- A small number of helpers are added to make generation of bits of SQL, such as a 'TableReferenceList' that includes the 'AS alias' piece.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/OrderBy.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/OrderBy.hs
@@ -27,7 +27,7 @@ where
 
 import qualified Data.List.NonEmpty as NEL
 
-import Orville.PostgreSQL.Expr.Name (Alias, ColumnName, Qualified)
+import Orville.PostgreSQL.Expr.Name (AliasExpr, ColumnName, Qualified)
 import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
 
 {- |
@@ -107,15 +107,15 @@ orderByColumnName :: Qualified ColumnName -> OrderByDirection -> OrderByExpr
 orderByColumnName =
   curry (orderByColumnsExpr . pure)
 
-{- | Create an 'OrderByExpr' for 'Alias' and 'OrderByDirection' pairs, ensuring commas as needed. For basic queries involving columns on some table(s), use 'orderByColumnsExpr'. This
+{- | Create an 'OrderByExpr' for 'AliasExpr' and 'OrderByDirection' pairs, ensuring commas as needed. For basic queries involving columns on some table(s), use 'orderByColumnsExpr'.
 
 @since 1.1.0.0
 -}
-orderByAliasesExpr :: NEL.NonEmpty (Alias, OrderByDirection) -> OrderByExpr
+orderByAliasesExpr :: NEL.NonEmpty (AliasExpr, OrderByDirection) -> OrderByExpr
 orderByAliasesExpr =
   OrderByExpr . RawSql.intercalate RawSql.commaSpace . fmap columnOrdering
  where
-  columnOrdering :: (Alias, OrderByDirection) -> RawSql.RawSql
+  columnOrdering :: (AliasExpr, OrderByDirection) -> RawSql.RawSql
   columnOrdering (alias, orderByDirection) =
     RawSql.toRawSql alias <> RawSql.space <> RawSql.toRawSql orderByDirection
 
@@ -124,7 +124,7 @@ orderByAliasesExpr =
 
 @since 1.1.0.0
 -}
-orderByAlias :: Alias -> OrderByDirection -> OrderByExpr
+orderByAlias :: AliasExpr -> OrderByDirection -> OrderByExpr
 orderByAlias =
   curry (orderByAliasesExpr . pure)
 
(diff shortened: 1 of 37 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/OrderBy.hs
```diff
@@ -83,23 +90,44 @@ appendOrderByExpr (OrderByExpr a) (OrderByExpr b) =
...
 
 @since 1.0.0.0
 -}
-orderByColumnName :: ColumnName -> OrderByDirection -> OrderByExpr
+orderByColumnName :: Qualified ColumnName -> OrderByDirection -> OrderByExpr
 orderByColumnName =
   curry (orderByColumnsExpr . pure)
 
+{- | Create an 'OrderByExpr' for 'Alias' and 'OrderByDirection' pairs, ensuring commas as needed. For basic queries involving columns on some table(s), use 'orderByColumnsExpr'. This
+
```
qxjit, line 111: Extraneous "This" on the end of this line?

orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -1135,4 +1114,64 @@ orderByField ::
...
+orderByAliasedField aliasedFieldDef =
+  Expr.orderByColumnName (fieldColumnName (getAlias aliasedFieldDef) (getFieldDefinition aliasedFieldDef))
+
+{- | A 'FieldDefinition' that might have been aliased. This is equivalent to a tuple of the field
+definition and 'Maybe Expr.Alias'.
+
+@since 1.1.0.0
+-}
+data AliasedFieldDefinition nullability a = AliasedFieldDefinition
+  { i_alias :: Maybe Expr.Alias
```
qxjit, line 1136: Why does `AliasedFieldDefinition` have a `Maybe Expr.Alias` rather than always having an alias?

orville-postgresql/src/Orville/PostgreSQL/Marshall/SqlComparable.hs
```diff
@@ -0,0 +1,296 @@
...
+-}
+greaterThanOrEqualTo :: SqlComparable a b => a -> b -> Expr.BooleanExpr
+greaterThanOrEqualTo a =
+  Expr.greaterThanOrEqualTo
+    (referenceValueExpression a)
+    . Expr.valueExpression
+    . toComparableSqlValue a
+
+-- like
+-- likeInsensitive
```
qxjit, line 186: Did you intend these as placeholders to implement later?

orville-postgresql/src/Orville/PostgreSQL/Marshall/SyntheticField.hs
```diff
@@ -32,11 +36,21 @@ import qualified Orville.PostgreSQL.Raw.SqlValue as SqlValue
 @since 1.0.0.0
 -}
 data SyntheticField a = SyntheticField
-  { _syntheticFieldExpression :: Expr.ValueExpression
-  , _syntheticFieldAlias :: FieldName
-  , _syntheticFieldValueFromSqlValue :: SqlValue.SqlValue -> Either String a
+  { i_syntheticFieldExpression :: Expr.ValueExpression
+  , i_syntheticFieldAlias :: FieldName
```
qxjit, line 40: Would it make sense for this alias field to be of type `Alias` now rather than `FieldName`?

orville-postgresql/src/Orville/PostgreSQL/Expr/OrderBy.hs
telser: Removed, thank you!

orville-postgresql/src/Orville/PostgreSQL/Marshall/SqlComparable.hs
telser: These were implemented already, stray comments were removed!

orville-postgresql/src/Orville/PostgreSQL/Marshall/SyntheticField.hs
telser: I've changed this with addition of the alias analog for `FieldName`.

qxjit (approved):

Issue #340, opened 2024-02-09
Row locking select options


aaronallen8455:
It would be nice to have `SelectOptions` support some of the row-level locking clauses that postgres offers, perhaps just `FOR UPDATE` if nothing else.

qxjit:
This is a great suggestion. I've added to our list of things to do before a next release.

telser:
@aaronallen8455 Support for row locking is added to selection options in c231aa12066e0a04b590b871dd768f1513ef591b. I think we should be good to close this issue, does that seem reasonable to you?

aaronallen8455:
Yes, thanks @telser!

PR #353, opened 2024-04-06
Adds a RowLockingClause to SelectOptions

telser:
This is to address #340.

Support for the various forms of locking in a select statements is added, as documented by PostgrSQL at:
https://www.postgresql.org/docs/12/sql-select.html#SQL-FOR-UPDATE-SHARE

This is another case of a select option being left preferred, much like limit.


qxjit:
@telser We had talked about adding more tests to this, but if that's not something we have time for right now maybe we should go ahead and merge this.

qxjit (approved):

PR #354, opened 2024-04-06
Bump dependency bounds and update which GHC are tested with

telser:
- Bump upper bounds on containers and time as reported by cabal outdated
- Minor version bumps in the 9.4 and 9.6 series
- Updates nightly to the latest, adding 9.8.2 to versions tested with
- Updates the cabal file with GHC versions tested against
- Removes a redundant constraint caught by newer versions of GHC.


ysangkok (approved):
seems to compile and also still compiles with `cabal test --prefer-oldest -w ghc-9.8.1`

onslaughtq (approved):

PR #355, opened 2024-04-26
Thread safe connections

aaronallen8455:
Executing multiple commands concurrently on the same connection results in one of those commands hanging indefinitely, possibly due to a dead-lock somewhere in libpq. This can occur when concurrency is used inside of a context where a single connection is reserved, i.e. transactions.

This change introduces an MVar that is used to serialize access to a connection for executing commands. The existing semantics around closing connections should be preserved through the use of a second independent MVar.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
@@ -34,12 +34,12 @@ module Orville.PostgreSQL.Raw.Connection
 where
 
 import Control.Concurrent (getNumCapabilities, threadWaitRead, threadWaitWrite)
-import Control.Concurrent.MVar (MVar, newMVar, tryReadMVar, tryTakeMVar)
+import Control.Concurrent.MVar (MVar, newMVar, tryTakeMVar, withMVar)
 import Control.Exception (Exception, mask, throwIO)
-import Control.Monad (void)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Builder as BSB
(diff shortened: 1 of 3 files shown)


orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
```diff
@@ -295,33 +319,34 @@ connect noticeReporting connString =
...
+    underlyingFinish :: (forall a. IO a -> IO a) -> IO ()
     underlyingFinish restore = do
-      underlyingConnection <- tryTakeMVar handle
-      restore (traverse LibPQ.finish underlyingConnection)
+      connCtx <- readIORef handle
+      mCloseLock <- tryTakeMVar $ i_connCloseLock connCtx
+      case (mCloseLock, i_connState connCtx) of
+        (Just (), OpenConnection underlyingConnection) -> do
+          writeIORef handle connCtx {i_connState = ClosedConnection}
+          restore (LibPQ.finish underlyingConnection)
```
aaronallen8455, line 340: I kept this as is but I don't know why unmasking occurs here.

telser:
This hasn't been forgotten! I think it deserves very careful consideration. 

The intent was definitely to have a connection per unit of concurrency. With the pool making it relatively easy to grab a new connection when starting a new thread. Definitely would like to understand if there are situations being hit where that is not possible or not reasonable. 

Nothing about this change jumps out as an issue. But again, being careful for the interface here seems worthwhile. Further, if the above intent is not clear then I'd say we need to address that **somehow**. Either via this or something else.

PR #358, opened 2024-05-29
Adds JoinExpr to model SQL JOIN

telser:
The pieces needed to represent @JOIN@ are now available at the 'expr' level.

- 'JoinType' represents the INNER/LEFT/RIGHT/LATERAL portion.

- 'JoinConstraint' represents the portion that would determine how to limit the result set. Currently only a helper for 'ON expr' is added. This is left general enough to support the 'USING(columnNames)' syntax as well in a future exercise.

- 'JoinExpr' represent the above pieces put together so that they can be easily folded together into a 'FromItemExpr'.

Perhaps the most controversial change included is renaming 'TableReferenceList' to 'FromItemExpr'. This change more closely matches the name in the PostgreSQL documentation. Which hopefully aids clarity as at least some users did not realize that the "Table" in that name was _not_ necessarily a user defined table like a 'TableDefinition', but could be something like a @JOIN@ or a subquery.


orville-postgresql/src/Orville/PostgreSQL/Expr/Query.hs
```diff
@@ -153,6 +157,43 @@ operatorSubquery querySpecificRawSql binOp valExpr (QueryExpr queryRawSql) =
...
+{- | Append the SQL @AS@ and given 'AliasExpr' to a 'QueryExpr'
+
+@since 1.1.0.0
+-}
+queryExprWithAlias :: AliasExpr -> QueryExpr -> QueryExpr
+queryExprWithAlias alias query =
+  QueryExpr $
+    (RawSql.parenthesized $ RawSql.toRawSql query)
+      <> RawSql.fromString " AS "
+      <> RawSql.toRawSql alias
```
jlavelle, line 169: It's interesting that someone could call `queryExprWithAlias "a1"` on their `QueryExpr` before passing it to `joinQueryExpr _ "a2" _ _`, which would result in an expression like `left join (...) as a1 as a2 on ...`.

telser: That's true, but there are also other silly things you can do with the expr api.

PR #363, opened 2024-06-14
Adds support for creating and migrating column comments

jlavelle:
Setting a `fieldDescription` on a `FieldDefinition` will now result in a SQL comment on that column.


orville-postgresql/src/Orville/PostgreSQL/Expr/Comment.hs
```diff
@@ -59,3 +60,14 @@ commentTableExpr tableName mbComment =
       , RawSql.fromString "IS"
       , maybe RawSql.nullLiteral RawSql.toRawSql mbComment
       ]
+
+commentColumnExpr :: Maybe SchemaName -> TableName -> ColumnName -> Maybe Comment -> CommentExpr
```
telser, line 64: Missing haddocks

telser (approved):

PR #365, opened 2024-08-07
Adds `planTraversable`, a generalization of `planList`

jlavelle:


ysangkok:
This would be nice when working with NonEmpty, which are pretty common. So I am a fan.

jbrechtel (approved):
:+1:

jbrechtel:
Maybe @telser or @onslaughtq should chime in before it's merged in the event we need docs updated or explicit test coverage for this.

jlavelle:
I added a test case

telser (approved):
Seems like a reasonable convenience function to me. 

Thanks for fixing the workflows too!

PR #369, opened 2024-08-16
Implements 'SqlComparable' instance for 'SqlMarshaller'

jlavelle:
I added the instance and a number of helper functions for using it, including `Plan`s like `findAllWhereByMarshaller`.

PostgreSQL does not support the input of anonymous row values, and the instance methods need to produce row `SqlValue`s and `ValueExpression`s. This meant that I had to add support for row values at the `SqlValue` level. The `SqlRowValue` constructor will produce a row of placeholders:

```sql
($1, $2, $3)
```

See the changes to `Test/RawSql.hs` for more examples.


telser (approved):
This seems like a really nice win.

PR #381, opened 2024-10-09
Add instance Semigroup SelectList

ysangkok:
Especially when doing `GROUP BY`, it would be nice to add derived and non-derived columns.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Query.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Query.hs
@@ -215,6 +215,11 @@ SQL.
 newtype SelectList = SelectList RawSql.RawSql
   deriving (RawSql.SqlExpression)
 
+-- | @since 1.1.0.0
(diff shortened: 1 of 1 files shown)


jlavelle (approved):
I like this idea. Perhaps we should add the `Monoid` instance `mempty = RawSql.Empty` so that it's easier to use `foldMap` to build a `SelectList`.

ysangkok:
It doesn't really make sense to have an empty `SelectList` though? So I'd prefer to do that in a separate PR.

jlavelle:
> It doesn't really make sense to have an empty `SelectList` though? So I'd prefer to do that in a separate PR.

I don't have a use case for an empty select list by itself, but `SELECT FROM table` is valid SQL, and you can make one today by passing an empty list to one of the functions in the `SelectList` module. The reason it's useful is specifically so that you can use `foldMap` or similar functions that require a `Monoid` constraint.

I'm fine with putting it in a separate PR anyway.

PR #382, opened 2024-10-20
Add more precise 'Qualified' tracking.

telser:
- Created a new 'QualifiedOrUnqualified' type to be precise in our meaning.

- Changed the 'Qualified' type to keep track of the qualification.

- Changed functions to handle or result in 'Qualified' when needed and 'QualifiedOrUnqualified' when not.

- Removed optionality of alias for qualification in a handful of places.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
@@ -14,6 +14,9 @@ module Orville.PostgreSQL.Expr.Internal.Name.Qualified
   , qualifyFunction
   , qualifyColumn
   , aliasQualifyColumn
+  , QualifiedOrUnqualified
+  , untrackQualified
+  , unqualified
   )
 where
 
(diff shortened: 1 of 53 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
```diff
@@ -97,38 +144,33 @@ reference is appropriate.
...
+  case mbSchemaName of
+    Nothing ->
+      Qualified
+        { i_qualifier = RawSql.toRawSql (toIdentifier tableName)
+        , i_itemQualified = RawSql.toRawSql (toIdentifier unqualifiedName)
+        }
+    Just schemaName ->
+      unsafeSchemaQualify schemaName
+        . RawSql.unsafeFromRawSql
+        $ RawSql.toRawSql (toIdentifier tableName) <> RawSql.dot <> RawSql.toRawSql (toIdentifier unqualifiedName)
```
jlavelle, line 156: The `TableName` qualifier ends up prepended to `i_itemQualified`, is that what we actually want? It could be an issue if we expose a function that gives access to the qualifier or the item at some point.

orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -1126,15 +1136,15 @@ orderByAliasedField ::
   Expr.OrderByDirection ->
   Expr.OrderByExpr
 orderByAliasedField aliasedFieldDef =
-  Expr.orderByColumnName (fieldColumnName (getAlias aliasedFieldDef) (getFieldDefinition aliasedFieldDef))
+  Expr.orderByColumnName (Expr.untrackQualified $ fieldAliasQualifiedColumnName (getAlias aliasedFieldDef) (getFieldDefinition aliasedFieldDef))
 
 {- | A 'FieldDefinition' that might have been aliased. This is equivalent to a tuple of the field
 definition and 'Maybe Expr.Alias'.
```
jlavelle, line 1142:
```suggestion
{- | A 'FieldDefinition' that has been aliased. This is equivalent to a tuple of the field
definition and 'Expr.Alias'.
```

jlavelle (approved):
Looks great overall, just a question about  `qualifyColumn` and an outdated doc comment.

orville-postgresql/src/Orville/PostgreSQL/Expr/Internal/Name/Qualified.hs
telser: That's a great point. I've updated now.

orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
telser: Fixed this, thank you!

PR #384, opened 2024-10-22
Adds COUNT support to aggregate and window expr

telser:
- Deprecates the previous bespoke functionality.

- Adds count functions much like avg, max, min, etc to both aggregate and window

- Adds functions with the same api as aggregate expr functions.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Aggregate.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Aggregate.hs
@@ -19,6 +19,10 @@ module Orville.PostgreSQL.Expr.Aggregate
   , bitXorAggregateFunction
   , boolAndAggregateFunction
   , boolOrAggregateFunction
+  , countExprAggregateFunction
+  , count1AggregateFunction
+  , countColumnAggregateFunction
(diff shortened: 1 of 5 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/Aggregate.hs
```diff
@@ -229,6 +233,49 @@ boolOrAggregateFunction ::
 boolOrAggregateFunction =
   singleParameterAggregateFunction Name.boolOrFunctionName
 
+{- | A simplified version of 'countAggregatFunction' for the SQL @count@ of a
```
jlavelle, line 236:
```suggestion
{- | A simplified version of 'countAggregateFunction' for the SQL @count@ of a
```
More instances of this in the file below as well.

jlavelle (approved):
Looks good, just a few typos in the doc comments.

orville-postgresql/src/Orville/PostgreSQL/Expr/Aggregate.hs
telser: <3 Thanks for catching this!

PR #386, opened 2024-10-23
Helper to use an `InsertSource` as a `QueryExpr`

jlavelle:


jlavelle:
I'm going to close this and add a proper `ValuesExpr` instead, since there are a number of other optional parts to a `VALUES` statement.

PR #387, opened 2024-10-24
Adds `ValuesExpr` for constructing `VALUES` SQL

jlavelle:
I added a new module for constructing `VALUES` statements and refactored `InsertSource` to use it. I deprecated the old `InsertSource` construction functions, which could be used to produce invalid SQL by passing empty lists of values.

This also had the effect of improving the behavior of `mkInsertSource` when called with a `SqlMarshaller` with no fields; it will now use `VALUES(DEFAULT)` for that case instead of generating invalid SQL.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Values.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Values.hs
@@ -1,0 +1,146 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{- |
+Copyright : Flipstone Technology Partners 2024
+License   : MIT
+Stability : Stable
+
+@since 1.1.0.0
+-}
+module Orville.PostgreSQL.Expr.Values
+  ( ValuesExpr
+  , valuesExpr
+  , valuesExprFromValueExpressions
+  , ValuesExprRow
+  , valuesExprRow
+  , ValuesExprValue
+  , valuesExprValue
+  , valuesExprDefaultValue
+  , valuesQueryExpr
+  ) where
+
+import Data.List.NonEmpty (NonEmpty)
+import Data.Maybe (catMaybes)
+
+import Orville.PostgreSQL.Expr.FetchClause (FetchClause)
+import Orville.PostgreSQL.Expr.LimitExpr (LimitExpr)
+import Orville.PostgreSQL.Expr.OffsetExpr (OffsetExpr)
+import Orville.PostgreSQL.Expr.OrderBy (OrderByExpr)
+import Orville.PostgreSQL.Expr.Query (QueryExpr)
+import Orville.PostgreSQL.Expr.ValueExpression (ValueExpression)
+import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
+
+{- |
+Type to represent a @VALUES@ statement, e.g.
+
+> VALUES ('Bob',32),('Cindy',33)
+
+@since 1.1.0.0
+-}
+newtype ValuesExpr = ValuesExpr RawSql.RawSql
+  deriving
+    ( -- | @since 1.1.0.0
+      RawSql.SqlExpression
+    )
+
+{- |
+  Construct a 'ValuesExpr' for the given 'ValuesExprRow's, and any of the supported
+  optional clauses.
+
+@since 1.1.0.0
+-}
+valuesExpr ::
+  NonEmpty ValuesExprRow ->
+  Maybe OrderByExpr ->
+  Maybe LimitExpr ->
+  Maybe OffsetExpr ->
+  Maybe FetchClause ->
+  ValuesExpr
+valuesExpr vals mbOrderBy mbLimit mbOffset mbFetch =
+  let
+    opts =
+      RawSql.intercalate RawSql.space $
+        catMaybes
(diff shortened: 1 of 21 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/Values.hs
```diff
@@ -0,0 +1,82 @@
...
+        <> opts
+
+{- |
+  Constructs the SQL @VALUES(DEFAULT)@. Use to insert a default row. Note that this cannot
+  be used as a valid query expression, unlike a 'ValuesExpr' constructed using 'valuesExpr'
+
+@since 1.1.0.0
+-}
+defaultValuesExpr :: ValuesExpr
+defaultValuesExpr = ValuesExpr $ RawSql.fromString "VALUES(DEFAULT)"
```
telser, line 74: Looking at https://www.postgresql.org/docs/13/sql-insert.html it seems like we can have `DEFAULT` and more expressions inside the parens here to set a default on only some columns.

Similarly, should we support the `DEFAULT VALUES` variant from that same page?

jlavelle: From https://www.postgresql.org/docs/current/sql-values.html:
> A constant or expression to compute and insert at the indicated place in the resulting table (set of rows). In a VALUES list appearing at the top level of an INSERT, an expression can be replaced by DEFAULT to indicate that the destination column's default value should be inserted. DEFAULT cannot be used when VALUES appears in other contexts.

I'm thinking of adding a `sqlDefault :: ValueExpression` (since `default` is keyword), then callers can pass that into the `valuesExpr` function. With that we could reimplement `defaultValuesExpr` as `valuesExpr (pure $ pure sqlDefault) ...`or remove it entirely. Thoughts?

jlavelle: Actually I'm not sure if it makes sense to have `DEFAULT` be a `ValueExpression` since it's not valid elsewhere. It's a little awkward to change `valuesExpr` to take a `Either Default ValueExpression` or something, though.

telser: I think we should have a new type for where we currently have `NonEmpty ValueExpression`. That would make it easier for the `valuesExpr` function because we'd move the `DEFAULT` portion to the creation of that type.

jlavelle: Addressed in https://github.com/flipstone/orville/pull/387/commits/01c41946e54948a44916947545a49ff544e2b3fc

orville-postgresql/src/Orville/PostgreSQL/Expr/Values.hs
```diff
@@ -0,0 +1,143 @@
...
+  Construct a 'ValuesExpr' for the given 'ValuesExprRow's, and any of the supported
+  optional clauses.
+
+@since 1.1.0.0
+-}
+valuesExpr ::
+  NonEmpty ValuesExprRow ->
+  Maybe OrderByExpr ->
+  Maybe LimitExpr ->
+  Maybe OffsetExpr ->
```
telser, line 55: Do we want to get #388 in first so we can support `FETCH` here without changing signature?

jlavelle: Yeah, that's a good idea. I just re-reviewed and approved #388 and can rebase this branch once we merge it.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -573,35 +572,53 @@ mkInsertColumnList marshaller =
...
-
-{- |
-  An internal helper function that collects the 'SqlValue' encoded value for a
-  field from a Haskell entity, adding it a list of 'SqlValue's that is being
-  built.
+    Expr.valuesExprInsertSource $
+      case traverse encodeRow entities of
+        Nothing ->
+          Expr.valuesExpr
+            (Expr.valuesExprRow (pure Expr.valuesExprDefaultValue) <$ entities)
```
telser, line 593: I think this is where we could use `DEFAULT VALUES` instead of `VALUES(DEFAULT)` right? 

Which as I understand would mean we don't have to mention the collection at all.

telser (approved):
Just a couple more questions/thoughts.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
jlavelle: That would only insert a single default row, instead of `length entities` default rows.

orville-postgresql/src/Orville/PostgreSQL/Expr/Values.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{- |
+Copyright : Flipstone Technology Partners 2023-2024
```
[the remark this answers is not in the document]
jlavelle: Thanks, fixed

PR #388, opened 2024-10-26
Add a 'FetchClause'

telser:
'FetchClause' is so named because it is not a top level item and to avoid the existing 'FetchExpr'.

Some of the function names have similar awkward naming conflicts with no obvious solution other than a big breaking change.

Extends 'SELECT' to support 'FETCH' using the new 'FetchClause'

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
@@ -1,0 +1,102 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{- |
+Copyright : Flipstone Technology Partners 2024
+License   : MIT
+Stability : Stable
+
+@since 1.1.0.0
+-}
+module Orville.PostgreSQL.Expr.FetchClause
+  ( FetchClause
+  , fetchClauseInt
+  , FetchClauseModifier
+  , withTiesFetchClauseModifier
+  , onlyFetchClauseModifier
+  )
+where
+
+import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
+import qualified Orville.PostgreSQL.Raw.SqlValue as SqlValue
+
+{- |
+Type to represent a SQL FETCH clause. E.G.
+
+> FETCH FIRST 2 ROWS ONLY
+
+'FetchClause' provides a 'RawSql.SqlExpression' instance. See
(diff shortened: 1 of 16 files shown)


orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
```diff
@@ -0,0 +1,98 @@
...
+newtype FetchClause
+  = FetchClause RawSql.RawSql
+  deriving
+    ( -- | @since 1.1.0.0
+      RawSql.SqlExpression
+    )
+
+fetchClauseInt :: Int -> FetchClauseModifier -> FetchClause
+fetchClauseInt fetchNumber =
+  internal_fetchRaw (RawSql.parameter (SqlValue.fromInt fetchNumber))
```
jlavelle, line 42: Missing docs

jlavelle (changes requested):

orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
```diff
@@ -0,0 +1,98 @@
...
+internal_fetchRaw :: RawSql.RawSql -> FetchClauseModifier -> FetchClause
+internal_fetchRaw countingRawSql modifier =
+  FetchClause $
+    RawSql.fromString "FETCH NEXT "
+      <> RawSql.parenthesized countingRawSql
+      <> RawSql.fromString " ROW "
+      <> RawSql.toRawSql modifier
+
+{- |
+Type to represent the treament of ties in a SQL FETCH expression. E.G.
```
ysangkok, line 67: should be `treatment`

orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
telser: Thank you! Fixed in [cd1df0a](https://github.com/flipstone/orville/pull/388/commits/cd1df0af9cb0d5e92bf0d576e9a5b0e40754d6b3)

orville-postgresql/src/Orville/PostgreSQL/Expr/FetchClause.hs
telser: Thanks! Fixed in [cd1df0a](https://github.com/flipstone/orville/pull/388/commits/cd1df0af9cb0d5e92bf0d576e9a5b0e40754d6b3)

jlavelle (approved):

PR #389, opened 2024-11-02
Extend 'UpdateExpr' with FROM and ONLY support

telser:
- 'FROM' allows for columns in other tables to be referenced in the 'SET' or 'WHERE' portions of the 'UPDATE'

- Support for the 'ONLY' keyword in 'UPDATE' allows for updating just the named table without updating tables inheriting from the named.

- Note that this does not support the `*`. Using `ONLY` and `*` is equivalent to not having either, so we support the range of functionality without all of the syntax.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/Update.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/Update.hs
@@ -14,14 +14,19 @@ module Orville.PostgreSQL.Expr.Update
   , setClauseList
   , SetClause
   , setColumn
+  , setColumnExpression
+  , UpdateNamedOnly
+  , onlyUpdateNamedTable
   )
 where
 
 import Data.List.NonEmpty (NonEmpty)
 import Data.Maybe (catMaybes)
 
+import Orville.PostgreSQL.Expr.FromItemExpr (FromItemExpr)
 import Orville.PostgreSQL.Expr.Name (ColumnName, QualifiedOrUnqualified, TableName)
 import Orville.PostgreSQL.Expr.ReturningExpr (ReturningExpr)
+import Orville.PostgreSQL.Expr.ValueExpression (ValueExpression)
 import Orville.PostgreSQL.Expr.WhereClause (WhereClause)
 import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
 import qualified Orville.PostgreSQL.Raw.SqlValue as SqlValue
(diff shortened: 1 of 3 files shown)


orville-postgresql/test/Test/Expr/InsertUpdateDelete.hs
```diff
@@ -352,3 +359,60 @@ prop_truncateTablesExpr =
...
+      expectedFooBars = NE.fromList [mkFooBar 1 "bird", mkFooBar 2 "cat", mkFooBar 3 "fish"]
+      fooBar2Table =
+        Expr.unqualified (Expr.tableName "foobar2")
+      dropAndRecreateSecondTable connection = do
+        RawSql.executeVoid connection (RawSql.fromString "DROP TABLE IF EXISTS " <> RawSql.toRawSql fooBar2Table)
+        RawSql.executeVoid connection (RawSql.fromString "CREATE TABLE " <> RawSql.toRawSql fooBar2Table <> RawSql.fromString "(foo INTEGER, bar TEXT)")
+
+      fooBar2FromItem = Expr.tableFromItemWithAlias (Expr.stringToAliasExpr "f2") fooBar2Table
+
+      setClauseList = RawSql.unsafeSqlExpression "bar = f2.bar"
```
jlavelle, line 378: We should probably add a helper (or change the existing function) to construct a `SetClauseList` that takes a `ValueExpression` instead of a `SqlValue` so that we can do this without `RawSql`.

orville-postgresql/src/Orville/PostgreSQL/Expr/Update.hs
```diff
@@ -47,30 +50,69 @@ newtype UpdateExpr
...
+        , fmap RawSql.toRawSql maybeUpdateNamedOnly
         , Just $ RawSql.toRawSql tableName
         , Just $ RawSql.fromString "SET"
         , Just $ RawSql.toRawSql setClause
+        , fmap buildFromItem maybeFromItemExpr
         , fmap RawSql.toRawSql maybeWhereClause
         , fmap RawSql.toRawSql maybeReturningExpr
         ]
 
+{- | Type to represent the optoin to update only the named table and not any descendants. E.G.
```
jlavelle, line 91:
```suggestion
{- | Type to represent the option to update only the named table and not any descendants. E.G.
```

jlavelle (approved):

orville-postgresql/test/Test/Expr/InsertUpdateDelete.hs
telser: I think we would want to do it at the individual `SetClause` level. I've added a function for this that better covers the the right side of the `=` being an expression. For this test though, going directly to the `SetClauseList` is more about expediency/brevity than anything else.

PR #391, opened 2024-11-09
Supports negative years for date and timestamp types

### LINKED ISSUE #380: 'Invalid syntax' generated with negative years
I noticed that if you try to insert a UTCTime with a negative year, it fails even though PostgreSQL does handle this case. It seems that we are generating invalid syntax.

On PostgresSQL 15, the error is:

```
SqlExecutionError
  { sqlExecutionErrorExecStatus = Just FatalError
  , sqlExecutionErrorMessage = "ERROR:  invalid input syntax for type timestamp with time zone: \"-0001-01-01 00:00:00+00\"\n"
  , sqlExecutionErrorSqlState = Just "22007"
  , sqlExecutionErrorSqlQuery = "INSERT INTO \"foo\" (\"id\",\"name\",\"birth\") VALUES ($1,$2,$3)"
  }
```

<details>
<summary>Reproduction program</summary>

```
import qualified Orville.PostgreSQL.AutoMigration as AutoMigration

import qualified Data.Int as Int
import qualified Data.Time as Time
import qualified Data.Text as T
type FooId = Int.Int32
type FooName = T.Text
type FooBirth = Time.UTCTime

data Foo = Foo
  { fooId :: FooId
  , fooName :: FooName
  , fooBirth :: FooBirth
  }
  deriving (Eq, Show)
fooIdField :: O.FieldDefinition O.NotNull FooId
fooIdField =
  O.integerField "id"

fooNameField :: O.FieldDefinition O.NotNull FooName
fooNameField =
  O.unboundedTextField "name"

fooBirthField :: O.FieldDefinition O.NotNull FooBirth
fooBirthField =
  O.utcTimestampField "birth"
fooMarshaller :: O.SqlMarshaller Foo Foo
fooMarshaller =
  Foo
    <$> O.marshallField fooId fooIdField
    <*> O.marshallField fooName fooNameField
    <*> O.marshallField fooBirth fooBirthField
table :: O.TableDefinition (O.HasKey FooId) Foo Foo
table =
  O.mkTableDefinition "foo" (O.primaryKey fooIdField) fooMarshaller
main :: IO ()
main = do
  pool <-
    O.createConnectionPool
        O.ConnectionOptions
          { O.connectionString = ""
          }

  mFoo <- O.runOrville pool $ do
    AutoMigration.autoMigrateSchema AutoMigration.defaultOptions [AutoMigration.SchemaTable table]
    _ <- O.deleteEntity table 0
    _ <- O.insertEntity table Foo { fooId = 0, fooName = T.pack "Name", fooBirth = Time.UTCTime (Time.fromGregorian (-1) 1 1) 0 }
    O.findEntity table 0
  print mFoo
```
</details>

I don't think we can actually map all UTCTimes to SQL, since PostgreSQL has a limit on the year, but the `time` library uses an `Integer`, which is unbounded.

One way to improve handling of UTCTime specifically, might be to use the function `make_timestamptz(year, month, ...)`, which explicitly states in its documentation that it supports negative years. Or maybe we just don't wanna support those? Either way, I think the choice should be documented.

So I propose we use Hedgehog to run property tests for all sqlTypes, to see which ones round trip and which ones don't. Then, we can attach documentation to the relevant SqlTypes and FieldDefinitions, with a comment warning that they are partial and will work for only a certain range. We can also test whether the ordering in SQL works the same as in Haskell.

jlavelle:
Previously we were ISO8601 encoding/decoding dates and timestamps with negative years, however PostgreSQL uses a "BC" suffix to indicate negative years. I added support for this syntax to the `Day`, `UTCTime`, and `LocalTime` `SqlType`s, and expanded the range of years generated in testing to ensure we're handling the negative years correctly.

Closes #380.

### DIFF
--- a/orville-postgresql/test/Test/PgGen.hs
+++ b/orville-postgresql/test/Test/PgGen.hs
@@ -125,7 +125,7 @@ pgLocalTime =
 
 pgDay :: HH.Gen Time.Day
 pgDay = do
-  year <- Gen.integral (Range.linearFrom 2000 0 3000)
+  year <- Gen.integral (Range.constantFrom 2000 (-4713) 294276)
   month <- Gen.integral (Range.constant 1 12)
   day <- Gen.integral (Range.constant 1 (Time.gregorianMonthLength year month))
 
--- a/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
@@ -608,6 +608,10 @@ jsonbField = fieldOfType SqlType.jsonb
 {- | Builds a 'FieldDefinition' that stores Haskell 'Time.Day' values as the
   PostgreSQL "DATE" type.
 
+  This field cannot represent the full range of 'Time.Day' values. PostgreSQL supports years
+  from -4731 to 5874897 inclusive for this field, and sending a 'Time.Day' with a year outside
+  of this range to the database will result in a PostgreSQL exception.
+
 @since 1.0.0.0
 -}
 dateField ::
@@ -619,6 +623,10 @@ dateField = fieldOfType SqlType.date
 {- | Builds a 'FieldDefinition' that stores Haskell 'Time.UTCTime' values as the
   PostgreSQL "TIMESTAMP with time zone" type.
 
+  This field cannot represent the full range of 'Time.UTCTime' values. PostgreSQL supports years
+  from -4731 to 294276 inclusive for this field, and sending a 'Time.UTCTime' with a year outside
+  of this range to the database will result in a PostgreSQL exception.
+
 @since 1.0.0.0
 -}
 utcTimestampField ::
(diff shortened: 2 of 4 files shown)


orville-postgresql/test/Test/PgGen.hs
```diff
@@ -125,7 +125,7 @@ pgLocalTime =
 
 pgDay :: HH.Gen Time.Day
 pgDay = do
-  year <- Gen.integral (Range.linearFrom 2000 0 3000)
+  year <- Gen.integral (Range.constant (-4713) 294276)
```
telser, line 128: Do we want to use [constantFrom](https://hackage.haskell.org/package/hedgehog-1.5/docs/Hedgehog-Range.html#v:constantFrom) to keep the origin at 2000?

telser (approved):
Looks good to me. Just a question for the tests.

orville-postgresql/test/Test/PgGen.hs
jlavelle: Did this in [c812edf](https://github.com/flipstone/orville/pull/391/commits/c812edf169d94cd0b56974c35b06e99537c336aa). It will make the generator shrink towards 2000, which I think is more reasonable and readable if there is a failure.

jlavelle:
There's an off-by-one bug because PostgreSQL considers `-1` to be 1 B.C.E, whereas time follows ISO8601 where `0` is 1 B.C.E. :)

jlavelle:
Closing for now because I have to make rather significant changes.

PR #393, opened 2024-11-26
Removes extra parens in `conflictTargetForIndexExpr`

jlavelle:
The `IndexBodyExpr` passed to `conflictTargetForIndexExpr` will already be parenthesized, and it is invalid PostgreSQL syntax to wrap it in an extra set of parens.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/OnConflict.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/OnConflict.hs
@@ -105,6 +105,7 @@ conflictTargetForIndexColumn colName mbWhere =
    for more than a single column. PostgreSQL will use the expression to infer which index to check for
    conflicts.
 
+   Note that this function assumes that the 'IndexBodyExpr' is parenthesized.
 @since 1.1.0.0
 -}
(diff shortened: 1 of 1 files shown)


onslaughtq (approved):

gera-cameron (approved):
~(looks (good (too (me))))~
looks good to me

orville-postgresql/src/Orville/PostgreSQL/Expr/OnConflict.hs
```diff
@@ -110,12 +110,12 @@ conflictTargetForIndexColumn colName mbWhere =
 conflictTargetForIndexExpr :: IndexBodyExpr -> Maybe WhereClause -> ConflictTargetExpr
 conflictTargetForIndexExpr idxBody mbWhere =
   let
-    parensBody = RawSql.parenthesized $ RawSql.toRawSql idxBody
```
telser, line 113: This is unfortunate. I see the helpers that are wrapping it in parens, but I don't think we can know for sure that _any_ `IndexBodyExpr` passed here will be wrapped that way (i.e. using the `SqlExpression` construction directly). :( That said, I _think_ you would want to put parens for then `IndexBodyExpr` always, so the assumption here is probably fine, but we should state that assumption in the docs.

telser (changes requested):

telser (approved):

PR #395, opened 2024-12-05
Improves transaction rollback handling

jlavelle:
I added a libpq transaction status check to the `BracketSuccess` branch of `withTransaction` that will ensure that the transaction has the `TransInTrans` state before issuing the success callback. If the status is `TransInError`, the transaction will be rolled back. If it is anything else, the transaction will be rolled back and an `UnexpectedTransactionStatusError` will be thrown.

Additionally, I added exception handling around the success SQL execution that will run the rollback callback if executing the SQL fails with a `SqlExecutionError`, which could happen if a transaction cannot be committed due to something like a serialization error.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
@@ -12,10 +12,13 @@ module Orville.PostgreSQL.Execution.Transaction
   ( withTransaction
   , inWithTransaction
   , InWithTransaction (InOutermostTransaction, InSavepointTransaction)
+  , UnexpectedTransactionStatusError (..)
   )
 where
 
+import Control.Exception (Exception, throwIO, try)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import qualified Database.PostgreSQL.LibPQ as LibPQ
 import Numeric.Natural (Natural)
 
 import qualified Orville.PostgreSQL.Execution.Execute as Execute
@@ -25,6 +28,7 @@ import qualified Orville.PostgreSQL.Internal.Bracket as Bracket
 import qualified Orville.PostgreSQL.Internal.MonadOrville as MonadOrville
 import qualified Orville.PostgreSQL.Internal.OrvilleState as OrvilleState
 import qualified Orville.PostgreSQL.Monad as Monad
+import qualified Orville.PostgreSQL.Raw.Connection as Connection
 import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
 
 {- | Performs an action in an Orville monad within a database transaction. The transaction
@@ -80,10 +84,29 @@ withTransaction action =
         liftIO $
           case result of
             Bracket.BracketSuccess -> do
-              let
-                successEvent = OrvilleState.transactionSuccessEvent transaction
-              executeTransactionSql (transactionEventSql state successEvent)
-              callback successEvent
+              mbTransactionStatus <- Connection.transactionStatus conn
+              case mbTransactionStatus of
+                Nothing -> do
+                  callback OrvilleState.RollbackTransaction
+                  throwIO $ UnexpectedTransactionStatusError Nothing
+                Just transactionStatus -> case transactionStatus of
+                  LibPQ.TransInTrans -> do
+                    let
+                      successEvent = OrvilleState.transactionSuccessEvent transaction
+                    eSuccess <- try $ executeTransactionSql (transactionEventSql state successEvent)
+                    case eSuccess of
+                      Right () ->
+                        callback successEvent
+                      Left ex -> do
+                        callback OrvilleState.RollbackTransaction
+                        throwIO (ex :: Connection.SqlExecutionError)
+                  LibPQ.TransInError -> do
+                    executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+                    callback OrvilleState.RollbackTransaction
+                  _ -> do
+                    executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+                    callback OrvilleState.RollbackTransaction
+                    throwIO $ UnexpectedTransactionStatusError (Just transactionStatus)
             Bracket.BracketError -> do
               let
                 rollbackEvent = OrvilleState.rollbackTransactionEvent transaction
@@ -91,6 +114,28 @@ withTransaction action =
               callback rollbackEvent
 
     Bracket.bracketWithResult beginTransaction finishTransaction doAction
+
+{- |
+  'withTransaction' will throw this exception if libpq reports a transaction status other
+  than 'LibPQ.TransInTrans' or 'LibPQ.TransInError', or if there is no connection to the
+  database. The latter case should be impossible, and indicates a bug in Orville if observed.
+
+@since 1.1.0.0
+-}
+newtype UnexpectedTransactionStatusError = UnexpectedTransactionStatusError
+  { unexpectedTransactionStatusErrorTransactionStatus :: Maybe LibPQ.TransactionStatus
+  }
+
+instance Show UnexpectedTransactionStatusError where
+  show =
+    maybe
+      "UnexpectedTransactionStatusError: No database connection."
+      ( \status ->
+          "UnexpectedTransactionStatusError: " <> show status
+      )
+      . unexpectedTransactionStatusErrorTransactionStatus
+
+instance Exception UnexpectedTransactionStatusError
 
 transactionEventSql ::
   OrvilleState.OrvilleState ->
--- a/orville-postgresql/test/Test/Transaction.hs
+++ b/orville-postgresql/test/Test/Transaction.hs
@@ -3,6 +3,7 @@ module Test.Transaction
   )
 where
 
+import Control.Exception (SomeException (..), catch)
 import qualified Control.Monad as Monad
 import qualified Data.ByteString as BS
 import qualified Data.IORef as IORef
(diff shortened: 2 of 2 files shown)


orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -76,22 +80,60 @@ withTransaction action =
...
                 successEvent = OrvilleState.transactionSuccessEvent transaction
-              executeTransactionSql (transactionEventSql state successEvent)
-              callback successEvent
+              eSuccess <- try $ executeTransactionSql (transactionEventSql state successEvent)
+              case eSuccess of
+                Right () ->
+                  callback successEvent
+                Left ex -> do
+                  callback OrvilleState.RollbackTransaction
+                  throwIO (ex :: Connection.SqlExecutionError)
```
jlavelle, line 99: This is probably incorrect if the error was from trying to release a savepoint, in which case we should run a savepoint rollback and call the corresponding callback.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -76,22 +80,60 @@ withTransaction action =
...
                 rollbackEvent = OrvilleState.rollbackTransactionEvent transaction
               executeTransactionSql (transactionEventSql state rollbackEvent)
               callback rollbackEvent
+          Just LibPQ.TransInError -> do
+            executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+            callback OrvilleState.RollbackTransaction
+          Just _ -> do
+            executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+            callback OrvilleState.RollbackTransaction
+            throwIO $ UnexpectedTransactionStatusError mbTransactionStatus
```
jlavelle, line 111: Similar to the above, we need to check to see if we're in a savepoint and run the savepoint rollback instead.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -76,22 +80,60 @@ withTransaction action =
...
+                  throwIO (ex :: Connection.SqlExecutionError)
             Bracket.BracketError -> do
               let
                 rollbackEvent = OrvilleState.rollbackTransactionEvent transaction
               executeTransactionSql (transactionEventSql state rollbackEvent)
               callback rollbackEvent
+          Just LibPQ.TransInError -> do
+            executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+            callback OrvilleState.RollbackTransaction
+          Just _ -> do
```
telser, line 108: One of the possible states is `TransActive` which at least according to [docs](https://www.stackage.org/haddock/lts-22.43/postgresql-libpq-0.10.0.0/Database-PostgreSQL-LibPQ.html#t:TransactionStatus). I haven't yet worked out if it is possible that we got here but the transaction would still be processing on the postgresql side. Things appear as though we should never reach that state because _I think_ everything we do ends up flowing through the C `PQexec` (by way of `LibPQ.exec` in Haskell) which waits for commands to "finish".

Similarly thinking about `TransIdle`, could we get that state if the raw sql executed included `BEGIN`/`COMMIT`? Again, I _think_ it would be an error, right? 

All of this said, I think we probably should handle the cases explicitly rather than with a wildcard so that we are forced to think through these things if/when the underlying type changes.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -76,22 +80,60 @@ withTransaction action =
           action
 
       finishTransaction :: MonadIO m => () -> Bracket.BracketResult -> m ()
-      finishTransaction () result =
-        liftIO $
-          case result of
+      finishTransaction () result = liftIO $ do
+        mbTransactionStatus <- Connection.transactionStatus conn
+        case mbTransactionStatus of
```
telser, line 85: How does this interact with the `OrvilleState.TransactionState`? We have some notion of an "outermost" transaction there, shouldn't we be able to take advantage of that here?

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
telser: Wouldn't the fact that we're in a savepoint mean that we must have been inside a transaction? If I'm understanding the `TransactionStatus` type correctly, I think at least one constructor could never have that happen, namely `TransIdle`. But I think we'd need to check in the `TransInError` case too, right?

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -76,22 +80,60 @@ withTransaction action =
...
+            executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+            callback OrvilleState.RollbackTransaction
+            throwIO $ UnexpectedTransactionStatusError mbTransactionStatus
 
     Bracket.bracketWithResult beginTransaction finishTransaction doAction
 
+{- |
+  'withTransaction' will throw this exception if libpq reports a transaction status other
+  than 'LibPQ.TransInTrans' or 'LibPQ.TransInError', or if there is no connection to the
+  database. The latter case should be impossible, and indicates a bug in Orville if observed.
```
ysangkok, line 118: You list three cases. It is unclear which one 'latter' refers to to, I typically see it in a context with only two cases.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
ysangkok: How would releasing a savepoint ever fail? If the connection dropped? In that case, how can the savepoint rollbar get run?

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jbrechtel: I think the idea is that if we're in a savepoint then we don't want to rollback the whole transaction but just release the save point we're in? That sounds like the right thing to me in general but probably not the source of any of the bugs we're seeing.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jlavelle: Releasing a savepoint fails if the transaction is invalid (the `TransInError` status from libpq), in which case we need to run `ROLLBACK TO SAVEPOINT` instead.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
ysangkok: @jbrechtel Doesn't releasing a savepoint mean that you internalize it into the existing transaction? That's how I read [the docs](https://www.postgresql.org/docs/current/sql-release-savepoint.html). Why would we do that when we encounter an error that isn't TransInError (which is what this branch is for?) I don't understand why an error condition in particular would be the point where we do this?

jlavelle: If we're in a savepoint and we get `TransInError` we need to rollback the savepoint. If we're in the wildcard branch we probably can't issue any more SQL commands on the connection, unless we're somehow in `TransIdle`, in which case something has gone terribly wrong and we're not in a transaction anyway. I think throwing and calling the rollback callback is the best we can do in that case.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jlavelle: Right, we shouldn't hit the `TransActive` state when using `PQexec`. I think for now it should be treated as an invariant violation if we hit it. I will explode out the wildcard branch to make things more explicit.

jbrechtel (commented):
I'm in favor of this change in general. I agree with @telser re: combining the concepts of Orville's transaction state and LibPQ's and handling each LibPQ state explicitly.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jbrechtel: > @jbrechtel Doesn't releasing a savepoint mean that you internalize it into the existing transaction?

You're right, I meant rollback not release. I hate "RELEASE" for savepoints to mean "committing" them. I'm always saying the wrong thing.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jlavelle: Yeah, we do take advantage of it indirectly via the events. I think that should be more clear now.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -64,35 +68,91 @@ withTransaction action =
...
+              OrvilleState.BeginTransaction ->
+                beginAction
+              _ ->
+                throwIO transactionError
+            LibPQ.TransInTrans -> case openEvent of
+              OrvilleState.NewSavepoint _ ->
+                beginAction
+              _ ->
+                throwIO transactionError
+            LibPQ.TransActive ->
```
telser, line 91: If we want to begin a transaction and one is already active, should this be starting a savepoint?

jlavelle: `TransActive` means that another command is currently executing on the connection, which shouldn't happen with Orville given the way we serialize access to the connection with an `MVar` and use the blocking `LibPQ.execParams` to run commands. The one that means we're already in a transaction is `TransInTrans` and savepoints are handled in that case.

telser: Oh right :facepalm:. Thank you for clearing it up.

telser (approved):
I think this is going to be a good backstop against connections being in unexpected states. Under usual circumstances that shouldn't happen, but great to have the handling.

PR #396, opened 2024-12-13
Enclose original value in NUL byte found exception

ysangkok:
If you get many null byte found exceptions, it's hard to find which values are
triggering it. Let's put the original value in there, then it might be easier
for the consumers.


telser (changes requested):
The desire to distinguish where the NUL byte is definitely real. But we made an intentional decision to not print raw values at the lowest level so that there was not a sensitive information leak. Since we can't know if the value is safe to print, we don't. 

As it is I don't think we'd want to do this, but I'm trying to think if there is a reasonable way to get the distinguishing of values and prevent the printing of potentially sensitive information.

PR #397, opened 2024-12-25
Add Full-Text Search Functions to_tsquery and to_tsvector to Orville

tusharad:
This PR introduces two new functions, toTSQuery and toTSVector, to the Orville library. These functions facilitate PostgreSQL full-text search capabilities by allowing users to create text search queries and vectors directly within orville. Both functions return ValueExpression, making them seamlessly integrable into selectClause and whereClause. I have added test case and some documentation for the TSVector module. 

Please review and let me know if there are any additional changes or enhancements required. Feedback and suggestions for improvement are welcome!

Thank you for your time and input!

@telser @jlavelle

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
@@ -1,0 +1,100 @@
+{- |
+Copyright : Flipstone Technology Partners 2024-2025
+License   : MIT
+Stability : Stable
+
+@since 1.1.0.0
+-}
+module Orville.PostgreSQL.Expr.TSVector (
+      toTSVector 
+    , tsMatch
+    , toTSQuery
+    , Regconfig(.. )   
+) where
+
+import Orville.PostgreSQL.Expr.Name (FunctionName, functionName)
+import Orville.PostgreSQL.Expr.BinaryOperator (BinaryOperator, binaryOperator, binaryOpExpression)
+import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
+import Orville.PostgreSQL.Expr.ValueExpression (ValueExpression, functionCall)
+import Orville.PostgreSQL.Expr.WhereClause (BooleanExpr)
+import Data.Char (toLower)
+
+{- | Represents a text search configuration. 
+
+ @since 1.1.0.0
+-}
+data Regconfig = Simple
+  | Arabic
+  | Armenian
+  | Basque
+  | Catalan
+  | Danish
+  | Dutch
+  | English
+  | Finnish
+  | French
+  | German
+  | Greek
+  | Hindi
+  | Hungarian
+  | Indonesian
+  | Irish
+  | Italian
+  | Lithuanian
+  | Nepali
+  | Norwegian
+  | Portuguese
+  | Romanian
+  | Russian
+  | Serbian
+  | Spanish
+  | Swedish
+  | Tamil
+  | Turkish
+  | Yiddish
+    deriving (Eq, Show)
+
+tsMatchOp :: BinaryOperator
+tsMatchOp = binaryOperator "@@"
+
+{- | Builds a ts_match boolean expression from two value expressions. Alias of @@ operator.
+
+@since 1.1.0.0
+-}
+tsMatch :: ValueExpression -> ValueExpression -> BooleanExpr
+tsMatch tsVector tsQuery = binaryOpExpression tsMatchOp tsVector tsQuery
+
+toTSVectorFunction :: FunctionName
+toTSVectorFunction = functionName "to_tsvector"
+
+toTSQueryFunction :: FunctionName
+toTSQueryFunction = functionName "to_tsquery"
+
+regconfigToValueExpression :: Regconfig -> ValueExpression
+regconfigToValueExpression regconfig = RawSql.unsafeFromRawSql $ (RawSql.fromString "\'" <>
+    (RawSql.fromString $ map toLower (show regconfig)) <>
+    RawSql.fromString "\'")
+
+
+{- | Builds a tsvector from a value expression. If a regconfig is provided, it will be used to build the tsvector. 
+
+@since 1.1.0.0
+-}
+toTSVector :: ValueExpression -> Maybe Regconfig -> ValueExpression
+toTSVector val mRegconfig = 
+    let valueExpressions = case mRegconfig of
+                                Nothing -> mempty
+                                Just regconfig -> [regconfigToValueExpression regconfig] 
+        in functionCall toTSVectorFunction (valueExpressions <> [val])
+
+
+{- | Builds a tsquery from a value expression. If a regconfig is provided, it will be used to build the tsvector. 
+
+@since 1.1.0.0
+-}
+toTSQuery :: ValueExpression -> Maybe Regconfig -> ValueExpression
+toTSQuery val mRegconfig = 
+    let valueExpressions = case mRegconfig of
+                                Nothing -> mempty
+                                Just regconfig -> [regconfigToValueExpression regconfig] 
+        in functionCall toTSQueryFunction (valueExpressions <> [val])
--- a/orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
 {- |
 Copyright : Flipstone Technology Partners 2024-2025
 License   : MIT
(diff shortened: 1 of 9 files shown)


telser (changes requested):
Thank you so much @tusharad! Really appreciate this! Particularly, it is fantastic that you included a test, haddocks and the since annotations! 

I have a few specific notes that I've left. In addition though, looking at https://www.postgresql.org/docs/13/functions-textsearch.html it seems we might want to have a `TSQuery` type, and maybe a `TSVector` type too. That would, I think, make supporting more of the text searching functionality a little more ergonomic at the cost of needing an explicit function to convert to a `ValueExpression`. 

To be more specific I'm thinking something that is roughly like(I didn't include _everything_ just what would be different/new):
```
newtype TSQuery = TSQuery RawSql
  deriving
    ( -- | @since 1.1.0.0
      RawSql.SqlExpression
    )


newtype TSVector = TSVector RawSql
  deriving
    ( -- | @since 1.1.0.0
      RawSql.SqlExpression
    )

-- Haddocks here should note that it is a requirement of the user to ensure the the words are combined by valid operators. 
toTSQuery :: ValueExpression -> Maybe Regconfig -> TSQuery

-- This doesn't seem to have the same limitation as the to_tsquery() function, but the haddocks probably should note the `ValueExpression` needs to result in text, not some other postgresql type. Also should call out that if the `ValueExpression` results in json or jsonb that the behavior is different.
toTSVector :: ValueExpression -> Maybe Regconfig -> TSVector

tsMatch :: TSQuery -> TSVector -> BooleanExpr

tsQueryToValueExpression :: TSQuery -> ValueExpression
tsVectorToValueExpression :: TSVector -> ValueExpression
```
Again, the point of the specific types is so we could, in the future, extend this with more of the functions around text search like implementing the the tsquery and tsvector versions of `||` without that being on just any `ValueExpression` as the behavior is different even among those two types.

I think there is some room to explore around the `tsMatch` there as what I propose does force an order. A couple of options might be:
- Provide both directions with different named functions
- Have an unsafe version that works on `ValueExpression` like there is now _and_ provide type specific versions 
- Maybe something else I haven't thought of?
Thoughts on this?

orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
```diff
@@ -0,0 +1,100 @@
...
+{- | Builds a tsvector from a value expression. If a regconfig is provided, it will be used to build the tsvector. 
+
+@since 1.1.0.0
+-}
+toTSVector :: ValueExpression -> Maybe Regconfig -> ValueExpression
+toTSVector val mRegconfig = 
+    let valueExpressions = case mRegconfig of
+                                Nothing -> mempty
+                                Just regconfig -> [regconfigToValueExpression regconfig] 
+        in functionCall toTSVectorFunction (valueExpressions <> [val])
```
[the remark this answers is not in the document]
telser: With my top level comment, you'd need something like `TSVector . RawSql.toRawSql $` at the beginning.

tusharad:
Thank you for the review and feedback!

Regarding the tsMatch options, I agree that having two separate functions like tsMatch and tsMatch_ seems reasonable. I also plan to add additional functions, such as setWeight and tsRank, as part of this work. I'll start incorporating the requested changes and improvements now and will update you once they're complete.

Thanks again for your input!

tusharad:
I've pushed the latest commit addressing the requested changes: 

- ✅ Added a newtype wrapper around `TSVector` and `TSQuery` for better type safety and clarity.  
- ✅ Introduced a newtype wrapper for `RegConfig` and included default configurations like `simpleRegconfig` and `englishRegconfig`.  
- ✅ Enhanced documentation for all exposed functions and types, including examples for Haddock.  
- ✅ Added new functions: `ts_rank` for ranking search results and `setweight` for weighting `TSVector` elements.  
- ✅ Included additional test cases. While they're not exhaustive, they ensure the new functionality is working as expected.  

**Doubts:** 
1. The current module name is `TSVector`. Would `TextSearch` be a more fitting name, given the expanded functionality? Open to suggestions on this.  
2. I’ve added the `@@` operator as an alias for tsMatch. I hope this approach aligns with the expectations.  considered adding an infix alias for tsVectorConcat. However, the `||` operator is already taken, and `.||` is in use by orExpr. Do you think an infix alias for tsVectorConcat is necessary? If so, what operator would you recommend we use?

Please review the changes and let me know if there’s anything else required. Thanks!

telser:
Thanks a ton @tusharad! I'll look at the code changes separately but wanted to address the questions below first/quickly.

> **Doubts:**
> 
> 1. The current module name is `TSVector`. Would `TextSearch` be a more fitting name, given the expanded functionality? Open to suggestions on this.

The naming is hard! I think calling the module `TextSearch` is probably a good idea. This functionality definitely belongs together. Finding the best presentation is the trick. Seems like a good idea to keep the name close to PostgreSQL even though one could reasonable call things like `pg_trgm` "text search" too.

> 2. I’ve added the `@@` operator as an alias for tsMatch. I hope this approach aligns with the expectations.  considered adding an infix alias for tsVectorConcat. However, the `||` operator is already taken, and `.||` is in use by orExpr. Do you think an infix alias for tsVectorConcat is necessary? If so, what operator would you recommend we use?
> 

My thoughts on this are that we want to provide a way to write code that is natural. Hopefully that is obvious. Part of that means not doing awkward things like qualifying operators and being consistent. Issues I see with the `||` operator:
1. `||` is _really_ overloaded. As you point out it is used for the logical OR. But it is also used in two different ways just for text search functionality per [this table](https://www.postgresql.org/docs/13/functions-textsearch.html#TEXTSEARCH-OPERATORS-TABLE). As you said, for the concatenation of `tsvector`, but also for doing an OR of `tsquery`. So we would need to come up with some alternative for _one_ of them and that is just a whole mess.

Similarly with the `@@`:
1. `@@` is overloaded too. A quick search led me to [geometric operators](https://www.postgresql.org/docs/13/functions-geometry.html#FUNCTIONS-GEOMETRY-OP-TABLE) which we obviously don't support _right now_, but we should consider it for future design.
2. `@@` has two directions, like `tsMatch` so we'd probably want _some_ pattern for the flipped version which deviates from the way you'd write the SQL by hand anyway and could lead to confusing errors if you write the flipped one but then see `@@` in the raw sql that is executed. That is really not good.

Given that we probably don't have a good path for `||`, I'd err on the side of not having `@@` and being consistent with the operator support for `TextSearch` overall. The same would seem to apply to the other operators in `TextSearch`, there are at least some clashes with [jsonb operators](https://www.postgresql.org/docs/13/functions-json.html#FUNCTIONS-JSONB-OP-TABLE) including another use of `@@`. This is leading me to think for some of these bigger buckets of functionality around specific types we might need to do a bigger code reorganization. For now though I think we should skip the operators to focus on basic features which the functions themselves provide.

orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
```diff
@@ -0,0 +1,382 @@
...
+{- | Type to represent a SQL value expression that evaluates to a TSRank. This could be a
+constant value, a column reference or any arbitrary calculated expression.
+
+'ValueExpression' provides a 'RawSql.SqlExpression' instance. See
+'RawSql.unsafeSqlExpression' for how to construct a value with your own custom
+SQL.
+
+ @since 1.1.0.0
+-}
+newtype TSRank
```
telser, line 85: `TSRank` is an interesting one. I can see how you'd end up creating the type when we have them for so many other things. However, I notice that it isn't an input anywhere. At least not in the same way that `TSVector` is. Seems like when we compute this it would always be, to PostgreSQL a `real` and thus usable anywhere a `real` is allowed. Is that right or am I missing a function either here or in the PostgreSQL docs?

Assuming that `TSRank` is _not_ an input to something else that can only work with it then I'm not sure we need it as a separate type. Compared to `RegConfig`, `TSVector`, `TSQuery` and `TSWeight` which are all restricted inputs to some other bit of functionality.

orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
```diff
@@ -0,0 +1,382 @@
...
+ @since 1.1.0.0
+-}
+newtype RegConfig
+  = RegConfig RawSql.RawSql
+  deriving
+    ( -- | @since 1.1.0.0
+      RawSql.SqlExpression
+    )
+
+{- | A newtype wrapper for a SQL expression that evaluates to a weight for a 'TSVector'.
```
telser, line 107:
```suggestion
{- | A type to represent the weight to be given to elements of a 'TSVector'

'TSWeight' provides a 'RawSql.SqlExpression' instance. See 'RawSql.unsafeSqlExpression' for how to construct a value with your own custom SQL.
```

orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
```diff
@@ -0,0 +1,382 @@
...
+    Filped version of 'tsMatch'
+
+@since 1.1.0.0
+-}
+tsMatch_ :: TSQuery -> TSVector -> BooleanExpr
+tsMatch_ tsQuery tsVector =
+  binaryOpExpression
+    tsMatchOp
+    (tsQueryToValueExpression tsQuery)
+    (tsVectorToValueExpression tsVector)
```
telser, line 150:
```suggestion
tsMatchTSQueryTSVector :: TSQuery -> TSVector -> BooleanExpr
tsMatchTSQueryTSVector = flip tsMatch
```

orville-postgresql/src/Orville/PostgreSQL/Expr/TSVector.hs
```diff
@@ -0,0 +1,382 @@
...
+@since 1.1.0.0
+-}
+tsWeightA :: TSWeight
+tsWeightA =
+  TSWeight $
+    RawSql.unsafeFromRawSql
+      ( RawSql.fromString "\'"
+          <> RawSql.fromString "A"
+          <> RawSql.fromString "\'"
+      )
```
telser, line 188:
```suggestion
  TSWeight $ RawSql.fromString "\'A\'"   
```

telser (changes requested):
This is coming together quite nicely. Had a handful more things to look at this round.

tusharad:
Hi @telser,  

Thank you for the thorough review and detailed feedback—it was incredibly helpful! I’ve made the following changes based on the input:  

- ✅ Renamed the `TSVector` module to `TextSearch` to better reflect the expanded scope of functionality.  
- ✅ Removed the `@@` operator alias and the `toRegconfig` function.  
- ✅ Fixed silly documentation mistakes that had slipped through earlier—apologies for those!  
- ✅ Removed the `TSRank` newtype, since it's not an input for anyone .  
- ✅ Ensured consistent usage of `regConfig`, adhering to the camelCase convention for clarity.  
- ✅ Updated the test cases to use `englishRegConfig`, which exercises the `regConfigToValueExpression` function.  

Please let me know if there’s anything else to address or improve. I appreciate your detailed suggestions and patience throughout this process!  

Thanks again for the review!

telser (approved):
This looks great! Thank you @tusharad.

PR #398, opened 2024-12-29
Add initial support for identity columns

telser:
- Expr support for column identity is added for columns that should be generated always or only by default

  - There is some oddity here. The specification for creating an identity column is to add a constraint. However when looking up in the catalog identity information is not held with other constraints.

- Field definitions can be marked as identity

- Support for identity at the field definition limits to non-nullable fields, as is the case in PostgreSQL.

- AutoMigration support for marking column identity

There is additional work needed for full support. This is left for the future. Particularly, identity columns can take options for the underlying sequence. However, supporting that will require being able to look up said sequence using a catalog table not currently supported: pg_depend.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
@@ -104,6 +104,10 @@ module Orville.PostgreSQL.Marshall.FieldDefinition
   , getFieldDefinition
   , getAlias
   , buildAliasedFieldDefinition
+  , FieldIdentityGeneration (GeneratedAlways, GeneratedByDefault)
+  , markAsIdentity
+  , unmarkIdentity
+  , fieldIdentityGeneration
   )
 where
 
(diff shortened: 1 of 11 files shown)


orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
```diff
@@ -141,6 +145,8 @@ data FieldDefinition nullability a = FieldDefinition
   , i_fieldDefaultValue :: Maybe (DefaultValue.DefaultValue a)
   , i_fieldDescription :: Maybe String
   , i_fieldTableConstraints :: [FieldName -> ConstraintDefinition.ConstraintDefinition]
+  , i_fieldIdentity :: IdentityGADT nullability
+  , i_fieldColumnConstraints :: [Expr.ColumnConstraint]
```
jlavelle, line 149: It looks like we never read this field and instead call the `fieldColumnConstraints` to compute the value. There is already some inconsistency in `fieldOfType`, where `i_fieldColumnConstraints` is initialized to `mempty`. I'm inclined to just remove the field since we're already using the computed value everywhere.

jlavelle (changes requested):
This looks great overall! I have a concern about the `i_fieldColumnConstraints` field.

orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
telser: Thank you! This is a great point. I had initially hoped we'd be able to keep something like `i_fieldColumnConstraints` and reuse it for future constraint work. That was before tying the identity to the nullability though which clearly removes the need for this field.

jlavelle (approved):

PR #403, opened 2025-04-17
Rolls back transactions when begin fails in `withTransaction`

jlavelle:
Previously Orville could leak a connection with an open transaction if the `BeginTransaction` callback failed, e.g. due to throwing an exception or being interrupted by an async exception. We added exception handling around the begin action and added a test that will fail if the exception is not handled correctly.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
@@ -16,7 +16,7 @@ module Orville.PostgreSQL.Execution.Transaction
   )
 where
 
-import Control.Exception (Exception, onException, throwIO, try)
+import Control.Exception (Exception, finally, onException, throwIO, try, uninterruptibleMask)
 import Control.Monad.IO.Class (liftIO)
 import qualified Database.PostgreSQL.LibPQ as LibPQ
 import Numeric.Natural (Natural)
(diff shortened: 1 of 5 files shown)


jbrechtel (approved):
You'll definitely want other eyes on this but this looks good to me.

jlavelle:
Marked as a draft because I'm going to add some additional exception safety

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -134,7 +137,10 @@ withTransaction action =
           LibPQ.TransUnknown ->
             throwIO transactionError
 
-    Bracket.bracketWithResult beginTransaction finishTransaction doAction
+    Bracket.bracketWithResult
```
telser, line 140: I like this change of shifting the liftIO until the end.

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
```diff
@@ -67,62 +67,63 @@ withTransaction action =
...
+            callback openEvent
+          transactionError = UnexpectedTransactionStatusError status openEvent
+        case status of
+          LibPQ.TransIdle -> case openEvent of
+            OrvilleState.BeginTransaction -> do
+              let
+                rollback = uninterruptibleMask $ \restore -> do
+                  executeTransactionSql (transactionEventSql state OrvilleState.RollbackTransaction)
+                  restore $ callback OrvilleState.RollbackTransaction
+              beginAction `onException` rollback
```
telser, line 86: I think we should check the transaction status again inside of here so that we don't issue a rollback without a corresponding begin.

telser (changes requested):

telser (approved):

orville-postgresql/src/Orville/PostgreSQL/Execution/Transaction.hs
jlavelle: Good point!

PR #404, opened 2025-05-07
Warn about short-circuiting monads

ysangkok:
Adds some documentation such that users understand that they can't use short-circuiting monads like ExceptT.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Internal/MonadOrville.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Internal/MonadOrville.hs
@@ -73,6 +73,9 @@ class
   implementation of the methods below for monads that implement
   'Control.Monad.IO.Unlift.MonadUnliftIO'.
 
+  Note that 'm' MUST NOT be short-circuiting, like e.g. MaybeT and ExceptT.
+  Transactions won't work right if it is.
+
 @since 1.0.0.0
 -}
 class MonadOrvilleControl m where
--- a/orville-postgresql/src/Orville/PostgreSQL/Internal/MonadOrville.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Internal/MonadOrville.hs
@@ -73,8 +73,18 @@ class
   implementation of the methods below for monads that implement
   'Control.Monad.IO.Unlift.MonadUnliftIO'.
 
-  Note that 'm' MUST NOT be short-circuiting, like e.g. MaybeT and ExceptT.
-  Transactions won't work right if it is.
+  Note: Orville assumes that execution of any monadic operations will continue
+  normally other than when an exception is thrown in the IO monad. The monad
+  implementing 'MonadOrvilleControl' allows for the monad execution to
+  terminate in other ways (such as a 'Nothing' in the case of 'MaybeT' or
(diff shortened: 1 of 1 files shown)


ysangkok:
What's going on with CI, @telser ?

orville-postgresql/src/Orville/PostgreSQL/Internal/MonadOrville.hs
```diff
@@ -73,6 +73,9 @@ class
   implementation of the methods below for monads that implement
   'Control.Monad.IO.Unlift.MonadUnliftIO'.
 
+  Note that 'm' MUST NOT be short-circuiting, as it is in e.g. Servant and
```
telser, line 76: This seems like we probably want to say a little more or a little less. If we call out two libraries specifically we might want to include information on safe usage. Or we might not want to list the particular libraries at all.

@qxjit and @jlavelle might have more to say here.

jlavelle: I don't think we should directly reference Servant and Happstack here. I don't have a great way to word it (though I'm happy to help put some more thought into it), but the problem is with functions like this implemented in terms of `MonadOrvilleControl`:
```haskell
f :: (MonadIO m, MonadOrvilleControl m) => m a -> m ()
f action = do 
  liftCatch catch (void action) (\(_ :: SomeException) -> pure ())
  liftIO $ putStrLn "done"
```
Orville assumes that that `f action` will always print "done," but if you have a monad with a short-circuiting bind or an early exit capability (e.g. `ContT` with `callCC`) then that assumption is false, since with monads like `ExceptT` you can pass in an action like `ExceptT . pure $ Left "oops"` and it won't possibly be able to reach the `putStrLn`.

Orville uses an internally implemented bracket to handle cleanup in `withTransaction`, and it has this problem.

ysangkok: @jlavelle @telser Ok, I have changed the wording to mention MaybeT and ExceptT instead of Happstack and Servant.

telser: This is still tricky IMO. Because we have used short-circuiting monads in the designated slot and it works out okay but only because they didn't _use_ the short circuit. I think it is correct to limit to not short circuiting monads because it avoids the issue entirely, but don't want to mislead others into thinking it can't work with things such as Servant when we have seen it be perfectly fine for years.

@qxjit Do you have thoughts on how to best word this?

jlavelle: Another thing I was thinking about was adding the bracket function as a class method to `MonadOrvilleControl`, because that would let users implement a version that actually did work with `ExceptT` and `MaybeT` (though still not for `ContT`).

ysangkok: I think it sounds sensible to check the status, and I am looking forward to the upcoming PR. I also think the proposed wording is fine. My main goal with this PR is just to provide a hint, such that users are not caught off guard.  That's why I am not very picky about the wording. I note that perfect is an enemy of the good. So I've adjusted the PR to contain the proposed wording.

qxjit (approved):

PR #405, opened 2025-05-08
Note unsupported values for partial SQLTypes

ysangkok:
Let's add some hints such that users don't get surprised that these SQLTypes do not roundtrip.


orville-postgresql/src/Orville/PostgreSQL/Marshall/SqlType.hs
```diff
@@ -368,6 +382,9 @@ timestampWithoutZone =
 {- |  'jsonb' represents any type that can be converted To and From JSON. This corresponds
    to the "JSONB" type in PostgreSQL.
 
+  Note that strings contained within must not contain NULL unicode codepoints,
+  as they are unsupported in PostgreSQL.
```
jlavelle, line 386: The others don't say anything about unicode so why don't we either remove it from here or add it everywhere else for consistency.

jlavelle (approved):
Good idea, found one slight inconsistency.

PR #406, opened 2025-06-04
Revises Alias and Join APIs

qxjit:
This revies the Alias and Join APIs to more closely reflect the SQL grammar. It also clarifies the usages of "alias" and "qualify" to ensure a consistent meaning is reflected in the Orville API.


telser (approved):
I'll do a pass to update years in the copyrights.

PR #407, opened 2025-06-15
Supports negative years in ISO style date/time format

### LINKED ISSUE #380: 'Invalid syntax' generated with negative years

jlavelle:
Previously the `PgTime` module supported a hybrid between ISO-8601 format date/times and PostgreSQL's ISO text format and would not parse or print negative years correctly.

I changed the module to handle negative `Day`, `LocalTime`, and `UTCTime`s by converting the years to and from the common era scheme and parsing or appending the string " BC" as necessary.

I also added documentation to the relevant `SqlType`s and `FieldDefinition`s describing PostgreSQL's bounds on years, and several tests that cover edge cases like BCE leap years.

Closes #380.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Marshall/FieldDefinition.hs
@@ -617,6 +617,10 @@ jsonbField = fieldOfType SqlType.jsonb
 {- | Builds a 'FieldDefinition' that stores Haskell 'Time.Day' values as the
   PostgreSQL "DATE" type.
 
+  This field cannot represent the full range of 'Time.Day' values. PostgreSQL supports years
+  from -4731 to 5874897 inclusive for this field, and sending a 'Time.Day' with a year outside
+  of this range to the database will result in a PostgreSQL exception.
+
 @since 1.0.0.0
 -}
 dateField ::
@@ -628,6 +632,10 @@ dateField = fieldOfType SqlType.date
 {- | Builds a 'FieldDefinition' that stores Haskell 'Time.UTCTime' values as the
   PostgreSQL "TIMESTAMP with time zone" type.
 
+  This field cannot represent the full range of 'Time.UTCTime' values. PostgreSQL supports years
+  from -4731 to 294276 inclusive for this field, and sending a 'Time.UTCTime' with a year outside
+  of this range to the database will result in a PostgreSQL exception.
+
 @since 1.0.0.0
 -}
 utcTimestampField ::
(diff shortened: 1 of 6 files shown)


qxjit (approved):
:heart:

PR #408, opened 2025-07-22
[Bugfix] Qualify IndexName for dropIndexExpr to fix automigration

azenna:
- Ran into an issue where automigration would attemnt to drop an index that belonged to a schema without qualifying it resulting in a SqlError
- `qualifyIndex` wrapper added
- `dropIndexExpr` api changed to take `QualifiedOrUnqualified IndexName`


qxjit:
Thank you! I'm built this successfully locally, so I'm going to merge it.

PR #422, opened 2025-12-16
Add more function support

OwenGraves:
This adds `mkFunction`, function argument support, some return types, auto-migration of functions before tables, and cascade support for dropping functions during migration.

These are all changes in preparation of supporting typeIds in a separate orville package.
https://github.com/jetify-com/typeid


orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionArgument.hs
```diff
@@ -0,0 +1,29 @@
...
+
+> CREATE FUNCTION my_function(x int, y int) RETURNS text
+
+Support for using this type in orville is currently very limited and experimental
+
+@since 1.2.0.0
+-}
+newtype FunctionArgument = FunctionArgument RawSql.RawSql
+  deriving
+    ( -- \| @since 1.2.0.0
```
telser: I don't think we need the \ here.

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionDefinition.hs
```diff
@@ -92,9 +121,33 @@ mkTriggerFunction ::
 mkTriggerFunction name language source =
   FunctionDefinition
```
telser: I think we should turn this into a call to `mkFunction` now.

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionArgument.hs
```diff
@@ -0,0 +1,29 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{- |
+Copyright : Flipstone Technology Partners 2025
+License   : MIT
+Stability : Stable
+
+@since 1.2.0.0
+-}
+module Orville.PostgreSQL.Schema.FunctionArgument
```
telser: This should not be defined here! Notice how you end up importing something from `Schema.*` in an `Expr.*` module, that should not happen. `Expr` should be the base layer.

orville-postgresql/package.yaml
telser, line 59: You missed adding your new module to the exposed ones here. In practice we'd expect most users to import functionality from Orville.PostgreSQL.Expr but for documentation generation I think we want to be sure it is included here.


```suggestion
    - Orville.PostgreSQL.Expr.Function
    - Orville.PostgreSQL.Expr.FunctionArgument
    
```

telser (changes requested):
A couple of changes, the biggest is the moving of the `FunctionArgument` module

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionDefinition.hs
OwenGraves: Sounds good, I've decided not to eta reduce it for clarity

orville-postgresql/src/Orville/PostgreSQL/Schema/FunctionArgument.hs
OwenGraves: Good call, I was thinking of Schema.FunctionIdentifier when I was initially adding this

telser (approved):
We've discussed out of band that we'll probably want to change something about the automigration situation before a release. I want there to be a public record of that at least somewhere though.

With that, I think this looks okay. Would have been nice if we saw the cascade/restrict support being used/hit in the tests somewhere. But I don't want to block for that at this moment.

PR #424, opened 2025-12-17
Add orville-postgresql-typeid package

OwenGraves:
This adds a new orville package that supports using `KindID`s in field definitions to implement [typeids](https://github.com/jetify-com/typeid) and generate them directly in postgres using [typeid-sql](https://github.com/jetify-com/typeid-sql).

---

Remaining items:
 - Confirm the licensing is correct
 - Confirm how we want to run this in CI. Are we okay dropping support for older ghc versions, the dependencies of mmzk-typeid conflict with older stack lts?
 - Confirm orville-postgresql version bounding in package.yaml. This should probably be 1.2.0.0 instead once we have a release, though it would be nice to avoid remembering to need to change this when that happens


orville-postgresql-typeid/src/Orville/TypeId/FieldDefinition.hs
```diff
@@ -0,0 +1,59 @@
...
+kindIDDefaultValue ::
+  forall prefix.
+  KnownSymbol prefix =>
+  Orville.DefaultValue (KindID.KindID prefix)
+kindIDDefaultValue =
+  Orville.rawSqlDefault
+    . RawSql.unsafeFromRawSql
+    -- See documentation on 'O.rawSqlDefault' if changing - this must exactly match @pg_get_expr@
+    $ RawSql.fromString "typeid_generate_text("
+      <> RawSql.stringLiteral (BS8.pack $ symbolVal' (proxy# @prefix))
```
jlavelle, line 37: Is there a specific reason we're using `proxy#` instead of `Proxy` from `Data.Proxy`?

OwenGraves: To appease Trevis 😛. But no, there's not a strong reason either way, but it's probably marginally more efficient guaranteeing no runtime overhead there

orville-postgresql-typeid/src/Orville/TypeId/FieldDefinition.hs
```diff
@@ -0,0 +1,59 @@
...
+Note that 'typeid_generate_text' (and its dependents) need to be loaded as postgres functions using Orville 'SchemaFunction'
+
+@since 0.1.0.0
+-}
+kindIDFieldDefinition ::
+  forall prefix.
+  KindIDClass.ValidPrefix prefix =>
+  String ->
+  Orville.FieldDefinition Orville.NotNull (KindID.KindID prefix)
+kindIDFieldDefinition fieldName =
```
qxjit: I think we might end up wanting 2 versions of this, on that has the default value attached and one that doesn't.

orville-postgresql-typeid/test/Test/Property.hs
```diff
@@ -0,0 +1,106 @@
...
+import qualified Hedgehog.Internal.Config as Config
+import qualified Hedgehog.Internal.Property as Property
+import qualified Hedgehog.Internal.Report as Report
+import qualified Hedgehog.Internal.Runner as Runner
+import qualified Hedgehog.Internal.Seed as Seed
+
+import qualified Orville.PostgreSQL as Orville
+
+type NamedProperty = (HH.PropertyName, HH.Property)
+type NamedDBProperty = Orville.ConnectionPool -> NamedProperty
```
qxjit, line 30: I this duplicating this is probably fine for now, but I will write something down about possibly having some sort of shared testing infrastructure code

qxjit (changes requested):
Just one minor addition about having field definitions with and with the default value attached

orville-postgresql-typeid/src/Orville/TypeId/FieldDefinition.hs
OwenGraves: Sounds good, I've added the two versions

qxjit (approved):

PR #426, opened 2026-01-05
Enable manually destroying connections

### LINKED ISSUE #425: Manually destroying connection pool
Is there any way to manually destroy the `ConnectionPool`? It is using `Data.Pool` under the hood, which [recommends calling `destroyAllResources`](https://hackage-content.haskell.org/package/resource-pool-0.5.0.0/docs/Data-Pool.html#v:newPool):

> Note: although the runtime system will destroy all idle resources when the pool is garbage collected, it's recommended to manually call [destroyAllResources](https://hackage.haskell.org/package/resource-pool-0.4.0.0/docs/Data-Pool.html#v:destroyAllResources) when you're done with the pool so that the resources are freed up as soon as possible.

However, there seems to be no way to do this with the current `orville-postgresql` interface (perhaps I'm missing something!), since the `ConnectionPool` constructor is not exported.

I'm thinking this could be enabled by either exporting the `ConnectionPool` constructor, providing a function of type `ConnectionPool -> Pool Connection`, or providing a function that directly destroys the connection pool without exposing the underlying implementation. I'm happy to draft a PR if the team is open to the idea.

ehofreiter:
Use `destroyAllResources` from `resource-pool`. Add the `resource-pool` package version where `destroyAllResources` was introduced as a minimum version constraint.

Fixes #425.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
@@ -22,7 +22,7 @@ module Orville.PostgreSQL.Raw.Connection
   , StripeOption (OneStripePerCapability, StripeCount)
   , ConnectionPool
   , createConnectionPool
-  , destroyAllConnections
+  , destroyIdleConnections
   , Connection
   , withPoolConnection
   , executeRaw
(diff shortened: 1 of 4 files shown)


orville-postgresql/src/Orville/PostgreSQL/Raw/Connection.hs
```diff
@@ -120,40 +121,52 @@ createConnectionPool options = do
       connPerStripe
 #endif
 
+{- | Destroy (close) all connections currently in the connection pool,
+proactively freeing the associated resources and returning the pool to an
+empty state. This can be useful when it is known that all connections have
+become invalid or are no longer needed.
+
+Connections in use, such as in 'withPoolConnection', are not considered /in/ the
+pool and will not be immediately affected by 'destroyAllConnections'.
```
telser: Is this true? So if a connection is active and a call to `destroyAllConnections` is made, then the connection remains valid after?

I just want to make sure we're clear to users here.

qxjit: Based on the tests it does seems like open connections in use are not affected. I didn't realize this when I suggested the name. Maybe `destroyIdleConnections` is more accurate?

@ehofreiter Would you mind renaming it and targeting the PR at the `release/1.1` branch instead of `main`? I'll merge the release branch back into `main` after cutting the new release.

telser: @ehofreiter I did the change target part of the above for you, so it's just the function name that is left.

qxjit: Ah, sorry. That's my own illiteracy. I wasn't sure if changing the target would rebase just the PR changes, or all the commits from `main` that aren't on the release branch. Now I know!

ehofreiter: Right, the function `destroyAllResources` only has access to destroy connections that are currently _in_ the pool, and acquiring a resource to use _removes_ the resource from the pool (and potentially returns it if possible after use).

I renamed the function to `destroyIdleConnections` as suggested and tweaked the comments and test descriptions accordingly in the latest commit.

qxjit: Thank you @ehofreiter! I'm out tomorrow, but I'll get this merged and released later this week (or possibly the weekend).

qxjit:
This has been released as [1.1.1.0](https://hackage.haskell.org/package/orville-postgresql-1.1.1.0)

PR #428, opened 2026-02-07
Adds upsert entity operations using `ON CONFLICT`

jlavelle:
Added variations of the `insert*` entity operations that use `ON CONFLICT <target> DO UPDATE` clauses to update values that conflict due to a constraint violation.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
@@ -22,6 +22,12 @@ module Orville.PostgreSQL.Execution.EntityOperations
   , updateFields
   , updateFieldsAndReturnEntities
   , updateFieldsAndReturnRowCount
+  , upsertEntity
+  , upsertEntityAndReturnRowCount
+  , upsertAndReturnEntity
+  , upsertEntities
+  , upsertEntitiesAndReturnRowCount
+  , upsertAndReturnEntities
   , deleteEntity
   , deleteEntityAndReturnRowCount
   , deleteAndReturnEntity
@@ -39,6 +45,7 @@ import Control.Exception (Exception, throwIO)
 import qualified Control.Monad as Monad
 import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Data.List.NonEmpty as NEL
 import Data.Maybe (listToMaybe)
 
 import qualified Orville.PostgreSQL.Execution.Delete as Delete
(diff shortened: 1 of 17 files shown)


qxjit (commented):
I think this is a great addition to the API. I think we should consider what the shape of the `Expr.ConflictTargetExpr` should be at the `EntityOperation` layer of the interface. In particular, I'd like it to be very straightforward for a new user to do an upsert by primary key without worrying about the the `Expr` API.

orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
```diff
@@ -419,3 +426,99 @@ instance Show EmptyUpdateError where
...
+
+@since 1.1.1.0.1
+-}
+upsertAndReturnEntities ::
+  Monad.MonadOrville m =>
+  Schema.TableDefinition key writeEntity readEntity ->
+  Expr.ConflictTargetExpr ->
+  NEL.NonEmpty writeEntity ->
+  m [readEntity]
+upsertAndReturnEntities tableDef target = do
```
telser: The `do` here is redundant

onslaughtq (dismissed):

jlavelle:
@qxjit This is updated to use a sum type for selecting the conflict target, so users can call it like `upsertEntity table ByPrimaryKey entity`

orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
```diff
@@ -419,3 +431,211 @@ instance Show EmptyUpdateError where
...
+  as a partial unique index.
+
+  @since 1.1.1.0.1
+  -}
+  ByConflictTargetExpr :: Expr.ConflictTargetExpr -> ConflictTarget
+
+{- |
+  An error resulting from attempting to construct an invalid 'Expr.ConflictTargetExpr'.
+
+  Maybe be thrown as an exception by the upsert functions if a valid 'Expr.ConflictTargetExpr' could not
```
qxjit, line 472: "Maybe be thrown" -> "Maybe be thrown"?

orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
```diff
@@ -419,3 +431,211 @@ instance Show EmptyUpdateError where
 
 -- | @since 1.0.0.0
 instance Exception EmptyUpdateError
+
+{- |
+  Specifies the target for the @ON CONFLICT@ clause of an upsert function.
+  See the [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) for more information.
+
+@since 1.1.1.0.1
```
qxjit, line 439: We've been putting `@since 1.2` on new functions since that will be the actual new release number once we release.

qxjit (approved):
This look great. I noticed one typo and I believe we decided to use the next release version in new haddocks, but please merge this whenever you're ready.

orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
jlavelle: Dang I forgot to do this before I merged!

PR #429, opened 2026-02-15
Use oldest release from each PostgreSQL major series

ysangkok:
Since it is more likely that users use the latest minor release, than a previous one, we should test Orville on the latest minor release from each major release series.

### DIFF
--- a/orville-postgresql-typeid/compose.yaml
+++ b/orville-postgresql-typeid/compose.yaml
@@ -23,31 +23,31 @@ services:
       IN_TOOLS_CONTAINER: "true"
 
   testdb-pg13:
-    image: postgres:13.23-alpine
+    image: postgres:13.0-alpine
     environment:
       POSTGRES_USER: orville_test
       POSTGRES_PASSWORD: orville
 
   testdb-pg14:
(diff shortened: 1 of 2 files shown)


telser:
I was thinking about this recently and I'm not sure that testing with the latest minor is actually what we want. Our SQL *should* work with any of the minor releases, though the PostgreSQL versioning [policy](https://www.postgresql.org/support/versioning/) isn't particularly strong in wording. So then perhaps we would want to test with the _first_ minor of each release to show that we don't _require_ a change introduced in a later minor release. A practical point in favor of this is less churn in trying to keep up with the minor versions across 5 major PostgreSQL versions.

The above isn't a hard rule, this is meant to be a conversation because I think it is reasonable to go either direction.

@qxjit do you have strong opinions in either direction here?

ysangkok:
@telser Ok, I have added a commit to use the first minor of each release instead!

qxjit:
Let's test against the first minor version of each release. Thank you @ysangkok

ysangkok:
Thanks for your comments. Is this good to merge? I ask because it isn't 'approved'.

telser (approved):

PR #430, opened 2026-04-07
Adds support for batched insert/upsert of multiple entities

qxjit:
This changes the signature of the entity operations that insert multiple
entities to allow callers to request batching be performed to insert
the values in multiple statements.

The most common use case is to insert many entities with many columns
that would exceed the postgresql parameter limit. This is supported
via auto batch sizing. Explicit batch sizing is support also.

The caller must also specify if Orville should do its inserts in a new
transaction or not, either because they do not want any transaction at
all or because the caller knows a transaction is already open and
does not want to incur the cost of an extra savepoint being created.

### DIFF
--- a/orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GADTs #-}
 
 {- |
@@ -18,6 +19,8 @@ module Orville.PostgreSQL.Execution.EntityOperations
   , insertEntities
   , insertAndReturnEntities
   , insertEntitiesAndReturnRowCount
+  , BatchInsertOption (InOneStatement, InBatches)
+  , BatchInsertTransactionality (WithNewTransaction, WithoutNewTransaction)
   , updateEntity
   , updateEntityAndReturnRowCount
   , updateAndReturnEntity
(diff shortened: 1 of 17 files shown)


orville-postgresql/src/Orville/PostgreSQL/Execution/EntityOperations.hs
```diff
@@ -143,11 +160,99 @@ insertEntitiesAndReturnRowCount tableDef =
...
+      combine initialAcc <$> runBatch (toUnbatched batchable)
+    InBatches batchSize transactionality ->
+      let
+        -- make sure to force the accumulator to WHNF at each batch
+        foldBatch !acc =
+          fmap (combine acc) . runBatch
+      in
+        withBatchInsertTransactionality
+          transactionality
+          (Monad.foldM foldBatch initialAcc . toBatched batchSize $ batchable)
```
telser: I noticed that `executeBatchOption` is _only_ ever used with 4 arguments directly, the last argument is applied after a function call or as part of composition. These use sites would however allow some specialization to happen, _if_ `executeBatchOption` is inlined (particularly where we have it with Int). But according to the [ghc users guide](https://downloads.haskell.org/ghc/9.10.3/docs/users_guide/exts/pragmas.html#pragma-INLINE)
> Moreover, GHC will only inline the function if it is fully applied, where “fully applied” means applied to as many arguments as appear (syntactically) on the LHS of the function definition.

So I think we could rewrite this as:


```suggestion
executeBatchOption batchOption runBatch initialAcc combine =
  case batchOption of
    InOneStatement ->
      fmap (combine initialAcc) . runBatch . toUnbatched 
    InBatches batchSize transactionality ->
      let
        -- make sure to force the accumulator to WHNF at each batch
        foldBatch !acc =
          fmap (combine acc) . runBatch
      in
        withBatchInsertTransactionality
          transactionality
          . Monad.foldM foldBatch initialAcc . toBatched batchSize 
```

Which, per the user's guide, would then allow function to be inlined and specialized. I think we could easily do the same to `executeBatchOption_`.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -516,15 +526,40 @@ mkInsertExpr returningOption tableDef maybeOnConflict entities =
...
+        maybeOnConflict
+        (mkTableReturningClause returningOption tableDef)
   in
-    Expr.insertExpr
-      (tableName tableDef)
-      (Just insertColumnList)
-      insertSource
-      maybeOnConflict
-      (mkTableReturningClause returningOption tableDef)
+    fmap mkExprForBatch batches
```
telser: Similar to my other comment I think we'd like for this function to be able to be inlined. It's a bit different since it is exported, but I think we could pretty easily modify the last line here to help calls that would otherwise by "fully applied" if the `entities` argument didn't appear on the LHS. The last line would become `fmap mkExprForBatch . batchNonEmpty autoBatchSize`.

orville-postgresql/test/Test/TableDefinition.hs
```diff
@@ -259,3 +275,99 @@ prop_insertDefaultValuesWithEmptyMarshaller =
...
+
+      onConflictWithParams =
+        mconcat
+          . replicate onConflictParamCount
+          $ RawSql.parameter (SqlValue.fromInt32 0)
+
+      onConflictExpr :: Expr.OnConflictExpr
+      onConflictExpr =
+        RawSql.unsafeFromRawSql
+          ( RawSql.fromString "ON CONFLICT DO NOTHING "
```
telser: Why not use `onConflictDoNothing` here?

telser (approved):
This looks good to me. Thanks for handling my early comments!

PR #431, opened 2026-06-12
Ports test suites to `tasty`

jlavelle:
This will allow us to do things like run individual tests and re-run with specific seeds more easily. I also found several singleton properties that needed multiple runs (e.g. the default value tests), and vice versa.

This turned up one bug: `integralDefault` used an `::integer` cast, which is too small. I changed it to use `::bigint` instead.


orville-postgresql/src/Orville/PostgreSQL/Marshall/DefaultValue.hs
```diff
@@ -74,7 +75,7 @@ integralDefault n =
       then
         DefaultValue . RawSql.unsafeFromRawSql $
           RawSql.stringLiteral decimalBytes
-            <> RawSql.fromString "::integer"
+            <> RawSql.fromString "::bigint"
```
telser, line 78: I think we have done ourselves a large disservice by abstracting this function in general. Because now if someone wants a default on something smaller than bigint, it could fail right? This makes me think we should have either a fast-follow or a split out commit where we either 
1. inline this back into the others and make sure we do the sql type cast explicitly for each
2. pass the typecast portion into the general function.

Maybe I'm wrong in that postgresql will always allow a default to be cast to something bigger? Even if so I think it would read really weird if looking at the database information.

orville-postgresql/test/Test/FieldDefinition.hs
```diff
@@ -28,27 +30,29 @@ import Test.Expr.TestSchema (sqlRowsToText)
...
-      <> uuidField pool
-      <> dateField pool
-      <> utcTimestampField pool
-      <> localTimestampField pool
-      <> jsonbField pool
-
-prop_userData :: Property.NamedProperty
+  Tasty.testGroup
+    "FieldDefinition"
+    [ TastyHH.testProperty "User data can be stored on FieldDefinition" prop_userData
```
telser, line 37: This is much nicer IMO

orville-postgresql/test/Test/FieldDefinition.hs
```diff
@@ -191,39 +195,34 @@ testFieldProperties ::
...
 testFieldProperties pool fieldDefName roundTripTest =
-  ( ( String.fromString (fieldDefName <> " - can round trip values (not null)")
-    , HH.property $ runRoundTripTest pool roundTripTest
+  Tasty.testGroup
+    fieldDefName
+    ( [ TastyHH.testProperty (fieldDefName <> " - can round trip values (not null)") (HH.property $ runRoundTripTest pool roundTripTest)
+      , TastyHH.testProperty (fieldDefName <> " - can round trip values (nullable)") (HH.property $ runNullableRoundTripTest pool roundTripTest)
+      , TastyHH.testProperty (fieldDefName <> " - cannot insert null values into a not null field") (Property.singletonProperty (runNullCounterExampleTest pool roundTripTest))
+      ]
+        ++ fmap (testDefaultValueProperties pool fieldDefName roundTripTest) (roundTripDefaultValueTests roundTripTest)
```
telser, line 206: <> instead of ++

telser (commented):
One nitpick and then a bigger question around what we should do about the bugfix. Otherwise looks great to me.

orville-postgresql/src/Orville/PostgreSQL/Marshall/DefaultValue.hs
jlavelle: I did some investigation and it's not quite this simple, unfortunately. We need to determine the cast based on the magnitude of the default value. I will add a follow-up commit to this PR to address the problem.

ysangkok (approved):

PR #432, opened 2026-06-21
Add table policy support

onslaughtq:
This commit adds support for adding and dropping polciies from a table. Additionally, you can set an attribute on tables to enable row-level security.

### DIFF
--- a/orville-postgresql/orville-postgresql.cabal
+++ b/orville-postgresql/orville-postgresql.cabal
@@ -1,11 +1,11 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.38.1.
+-- This file has been generated from package.yaml by hpack version 0.39.6.
 --
 -- see: https://github.com/sol/hpack
 
 name:           orville-postgresql
-version:        1.1.1.0.1
+version:        1.1.1.0.2
 synopsis:       A Haskell library for PostgreSQL
 description:    Orville's goal is to provide a powerful API for applications to access PostgreSQL databases with minimal use of sophisticated language techniques or extensions. See https://github.com/flipstone/orville for more details.
 category:       database, library, postgresql
@@ -146,6 +146,7 @@ library
       Orville.PostgreSQL.Expr.Internal.Name.FunctionName
       Orville.PostgreSQL.Expr.Internal.Name.Identifier
       Orville.PostgreSQL.Expr.Internal.Name.IndexName
+      Orville.PostgreSQL.Expr.Internal.Name.PolicyName
       Orville.PostgreSQL.Expr.Internal.Name.Qualified
       Orville.PostgreSQL.Expr.Internal.Name.SavepointName
       Orville.PostgreSQL.Expr.Internal.Name.SchemaName
@@ -153,6 +154,10 @@ library
       Orville.PostgreSQL.Expr.Internal.Name.TableName
       Orville.PostgreSQL.Expr.Internal.Name.TriggerName
       Orville.PostgreSQL.Expr.Internal.Name.WindowName
+      Orville.PostgreSQL.Expr.PolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.CreatePolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.DropPolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.PolicyRole
       Orville.PostgreSQL.Internal.Bracket
       Orville.PostgreSQL.Internal.Extra.NonEmpty
       Orville.PostgreSQL.Internal.FieldName
@@ -176,6 +181,7 @@ library
       Orville.PostgreSQL.PgCatalog.PgSequence
       Orville.PostgreSQL.PgCatalog.PgTrigger
       Orville.PostgreSQL.Schema.ConstraintIdentifier
+      Orville.PostgreSQL.Schema.PolicyDefinition
       Paths_orville_postgresql
   hs-source-dirs:
       src
--- a/orville-postgresql/orville-postgresql.cabal
+++ b/orville-postgresql/orville-postgresql.cabal
@@ -76,6 +76,7 @@ library
       Orville.PostgreSQL.Expr.OnConflict
       Orville.PostgreSQL.Expr.OrReplace
       Orville.PostgreSQL.Expr.OrderBy
+      Orville.PostgreSQL.Expr.PolicyExpr
       Orville.PostgreSQL.Expr.Query
       Orville.PostgreSQL.Expr.ReturningExpr
       Orville.PostgreSQL.Expr.RowLocking
@@ -130,6 +131,7 @@ library
       Orville.PostgreSQL.Schema.FunctionDefinition
       Orville.PostgreSQL.Schema.FunctionIdentifier
       Orville.PostgreSQL.Schema.IndexDefinition
+      Orville.PostgreSQL.Schema.PolicyDefinition
       Orville.PostgreSQL.Schema.PrimaryKey
       Orville.PostgreSQL.Schema.SequenceDefinition
       Orville.PostgreSQL.Schema.SequenceIdentifier
@@ -154,7 +156,6 @@ library
       Orville.PostgreSQL.Expr.Internal.Name.TableName
       Orville.PostgreSQL.Expr.Internal.Name.TriggerName
       Orville.PostgreSQL.Expr.Internal.Name.WindowName
-      Orville.PostgreSQL.Expr.PolicyExpr
       Orville.PostgreSQL.Expr.PolicyExpr.CreatePolicyExpr
       Orville.PostgreSQL.Expr.PolicyExpr.DropPolicyExpr
       Orville.PostgreSQL.Expr.PolicyExpr.PolicyRole
@@ -181,7 +182,6 @@ library
       Orville.PostgreSQL.PgCatalog.PgSequence
       Orville.PostgreSQL.PgCatalog.PgTrigger
       Orville.PostgreSQL.Schema.ConstraintIdentifier
-      Orville.PostgreSQL.Schema.PolicyDefinition
       Paths_orville_postgresql
   hs-source-dirs:
       src
--- a/orville-postgresql/orville-postgresql.cabal
+++ b/orville-postgresql/orville-postgresql.cabal
@@ -166,6 +166,7 @@ library
       Orville.PostgreSQL.Internal.MigrationLock
       Orville.PostgreSQL.Internal.MonadOrville
       Orville.PostgreSQL.Internal.OrvilleState
+      Orville.PostgreSQL.Internal.PgArrayText
       Orville.PostgreSQL.Internal.RowCountExpectation
       Orville.PostgreSQL.Internal.TypeMap
       Orville.PostgreSQL.PgCatalog.DatabaseDescription
@@ -178,6 +179,7 @@ library
       Orville.PostgreSQL.PgCatalog.PgExtension
       Orville.PostgreSQL.PgCatalog.PgIndex
       Orville.PostgreSQL.PgCatalog.PgNamespace
+      Orville.PostgreSQL.PgCatalog.PgPolicy
       Orville.PostgreSQL.PgCatalog.PgProc
       Orville.PostgreSQL.PgCatalog.PgSequence
       Orville.PostgreSQL.PgCatalog.PgTrigger
--- a/orville-postgresql/orville-postgresql.cabal
+++ b/orville-postgresql/orville-postgresql.cabal
@@ -237,6 +237,7 @@ test-suite spec
       Test.Expr.Join
       Test.Expr.Math
       Test.Expr.OrderBy
+      Test.Expr.Policy
       Test.Expr.SequenceDefinition
       Test.Expr.TableDefinition
       Test.Expr.TableReferenceList
--- a/orville-postgresql/orville-postgresql.cabal
+++ b/orville-postgresql/orville-postgresql.cabal
@@ -77,6 +77,9 @@ library
       Orville.PostgreSQL.Expr.OrReplace
       Orville.PostgreSQL.Expr.OrderBy
       Orville.PostgreSQL.Expr.PolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.CreatePolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.DropPolicyExpr
+      Orville.PostgreSQL.Expr.PolicyExpr.PolicyRole
       Orville.PostgreSQL.Expr.Query
       Orville.PostgreSQL.Expr.ReturningExpr
       Orville.PostgreSQL.Expr.RowLocking
@@ -156,9 +159,6 @@ library
       Orville.PostgreSQL.Expr.Internal.Name.TableName
       Orville.PostgreSQL.Expr.Internal.Name.TriggerName
       Orville.PostgreSQL.Expr.Internal.Name.WindowName
-      Orville.PostgreSQL.Expr.PolicyExpr.CreatePolicyExpr
-      Orville.PostgreSQL.Expr.PolicyExpr.DropPolicyExpr
-      Orville.PostgreSQL.Expr.PolicyExpr.PolicyRole
       Orville.PostgreSQL.Internal.Bracket
       Orville.PostgreSQL.Internal.Extra.NonEmpty
       Orville.PostgreSQL.Internal.FieldName
--- a/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
+++ b/orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
@@ -3,7 +3,7 @@
 {-# LANGUAGE RankNTypes #-}
 
 {- |
-Copyright : Flipstone Technology Partners 2023-2025
+Copyright : Flipstone Technology Partners 2023-2026
 License   : MIT
 Stability : Stable
 
@@ -241,6 +241,7 @@ mkMigrationStepWithType stepType sql =
 isMigrationStepTransactional :: MigrationStepWithType -> Bool
 isMigrationStepTransactional stepWithType =
   case migrationStepType stepWithType of
+    DropPolicyDefinitions -> True
     DropForeignKeys -> True
     DropConstraints -> True
     DropIndexes -> True
@@ -257,6 +258,7 @@ isMigrationStepTransactional stepWithType =
     AddForeignKeys -> True
     AddIndexesConcurrently -> False
     SetComments -> True
+    AddPolicyDefinitions -> True
 
 {- | Indicates the kind of operation being performed by a 'MigrationStep' so
   that the steps can be ordered in a sequence that is guaranteed to succeed.
@@ -266,7 +268,9 @@ isMigrationStepTransactional stepWithType =
 @since 1.0.0.0
 -}
 data StepType
-  = -- | @since 1.1.0.0.3
+  = -- | @since 1.2.0.0
+    DropPolicyDefinitions
+  | -- | @since 1.1.0.0.3
     DropForeignKeys -- Foreign key constraints must be dropped before unique constraints
   | -- | @since 1.1.0.0.3
     DropConstraints
@@ -298,6 +302,8 @@ data StepType
     AddIndexesConcurrently
   | -- | @since 1.1.0.0
     SetComments
+  | -- | @since 1.2.0.0
+    AddPolicyDefinitions
   deriving
     ( -- | @since 1.0.0.0
       Eq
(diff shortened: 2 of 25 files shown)


orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -642,3 +658,75 @@ collectSqlValue entry encodeRest entity =
       encodeRest entity
     Synthetic _ ->
       encodeRest entity
+
+{- | Retrieves all the table policy definitions that have been added to the table via
+  'addTablePolicyDefinitions'.
+
```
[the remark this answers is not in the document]
onslaughtq: Fixed.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -362,7 +378,7 @@ tableTriggers =
 
 {- | Adds the given table triggers to the table definition.
 
-  Note: If multiple wriggers are added with the same 'Expr.TriggerName', only
+  Note: If multiple triggers are added with the same 'Expr.TriggerName', only
   the last one that is added will be part of the 'TableDefinition'. Any
   previously-added constraint with the same key is replaced by the new one.
```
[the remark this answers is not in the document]
onslaughtq: Fixed.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -642,3 +658,75 @@ collectSqlValue entry encodeRest entity =
...
+@since 1.2.0.0
+-}
+addTablePolicies ::
+  [PolicyDefinition] ->
+  TableDefinition key writeEntity readEntity ->
+  TableDefinition key writeEntity readEntity
+addTablePolicies policyDefs tableDef =
+  tableDef
+    { i_tablePolicyDefinitions = Set.fromList policyDefs
+    }
```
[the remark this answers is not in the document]
onslaughtq: My plan right now is to make the policies a Map like indices and triggers.

orville-postgresql/src/Orville/PostgreSQL/Expr/PolicyExpr/CreatePolicyExpr.hs
```diff
@@ -0,0 +1,237 @@
...
+    )
+
+-- | @since 1.2.0.0
+instance Eq PolicyUsingExpr where
+  (==) = on (==) (BS.map toLowerWord8 . RawSql.toExampleBytes)
+
+-- | @since 1.2.0.0
+instance Ord PolicyUsingExpr where
+  compare = on compare (BS.map toLowerWord8 . RawSql.toExampleBytes)
+
```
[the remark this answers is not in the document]
onslaughtq: I need to look into this more, since I'm not certain how postgresql deals with case-sensitivity of quoted identifiers. I'm thinking this isn't a real issue right now.

orville-postgresql/src/Orville/PostgreSQL/Expr/PolicyExpr/PolicyRole.hs
```diff
@@ -0,0 +1,46 @@
...
+import qualified Orville.PostgreSQL.Raw.RawSql as RawSql
+
+{- | Type to represent the name of a database role that a policy applies to, as
+used in the @TO@ clause of a @CREATE POLICY@ statement. 'PolicyRole' values
+constructed via the 'policyRole' function will be properly escaped as part of
+the generated SQL. E.G.
+
+> "some_role"
+
+'PolicyRole' provides a 'RawSql.SqlExpression' instance. See
```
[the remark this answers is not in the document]
onslaughtq: I'll fix this either in this PR or a follow-up PR, depending on if we want to merge this as-is to unblock some other stuff

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -677,12 +696,27 @@ mkCreateTableSteps currentNamespace tableDef =
...
         (mkAddTriggerSteps Set.empty tableName)
         (Orville.tableTriggers tableDef)
+
+    desiredPolicies = Orville.tablePolicies tableDef
+    wantsRLS = Orville.tableRowLevelSecurity tableDef
+
+    enableRLSSteps =
+      if wantsRLS && not (Set.null desiredPolicies)
+        then [mkMigrationStepWithType AddPolicyDefinitions (Expr.enableRowLevelSecurityExpr tableName)]
+        else []
```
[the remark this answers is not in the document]
onslaughtq: I'll fix this either in this PR or a follow-up PR, depending on if we want to merge this as-is to unblock some other stuff

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -833,6 +868,43 @@ mkAlterTableSteps currentNamespace relationDesc tableDef =
...
+      [ mkMigrationStepWithType DropPolicyDefinitions (Schema.mkDropPolicyExpr tableId existingPd)
+      | existingPd <- Set.toList existing
+      , Set.member (Schema.policyDefinitionPolicyName existingPd) polsToDrop
+      ]
+
+    -- Enable RLS if the table should have it and we're adding new policies
+    enableRLSSteps =
+      if wantsRLS && not (null createPolicySteps)
+        then [mkMigrationStepWithType AddPolicyDefinitions (Expr.enableRowLevelSecurityExpr tableName)]
+        else []
```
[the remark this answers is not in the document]
onslaughtq: I'll fix this either in this PR or a follow-up PR, depending on if we want to merge this as-is to unblock some other stuff.
I'm thinking though that enabling RLS shouldn't actually check for create/alter policy steps at all and be independent.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1749,128 @@ findCurrentNamespace = do
...
+    case tableNames of
+      [] -> pure Map.empty
+      (firstName : restNames) -> do
+        let
+          nonEmptyTableNames = firstName :| restNames
+          schemaNameStr = PgCatalog.namespaceNameToString currentNamespace
+          whereCondition =
+            Orville.fieldIn pgPolicyTableNameField nonEmptyTableNames
+              `Orville.andExpr` Orville.fieldEquals pgPolicySchemaNameField (T.pack schemaNameStr)
+
```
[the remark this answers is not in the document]
onslaughtq: I'll fix this either in this PR or a follow-up PR, depending on if we want to merge this as-is to unblock some other stuff

orville-postgresql/src/Orville/PostgreSQL/Expr/PolicyExpr/CreatePolicyExpr.hs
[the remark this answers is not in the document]
jlavelle: Quoted identifiers are case-sensitive in PostgreSQL. This instance is wrong, I think.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -512,12 +520,20 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
...
-              mkAlterTableSteps currentNamespace relationDesc tableDef
+              mkAlterTableSteps currentNamespace existingPolicies relationDesc tableDef
     SchemaDropTable tableId ->
       Right $
         let
           (schemaName, tableName) =
             tableIdToPgCatalogNames currentNamespace tableId
+
+          qualifiedTableId = setSchemaNameOnTableId currentNamespace tableId
+          existing = Maybe.fromMaybe Set.empty $ Map.lookup qualifiedTableId existingPolicies
```
telser: We should use Map.findWithDefault instead of the lookup with fromMaybe.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -512,12 +520,20 @@ calculateMigrationSteps currentNamespace dbDesc schemaItem =
...
+
+          dropPolicySteps =
+            [ mkMigrationStepWithType DropPolicyDefinitions (Schema.mkDropPolicyExpr qualifiedTableId pd)
+            | pd <- Set.toList existing
+            ]
```
telser:
```suggestion
          dropPolicySteps = fmap mkStep $ Set.toList existing
          mkStep pd =
            mkMigrationStepWithType DropPolicyDefinitions (Schema.mkDropPolicyExpr qualifiedTableId pd)
```

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -677,12 +696,27 @@ mkCreateTableSteps currentNamespace tableDef =
...
+
+    enableRLSSteps =
+      if wantsRLS && not (Set.null desiredPolicies)
+        then [mkMigrationStepWithType AddPolicyDefinitions (Expr.enableRowLevelSecurityExpr tableName)]
+        else []
+
+    addPolicySteps =
+      [ mkMigrationStepWithType AddPolicyDefinitions (Schema.mkCreatePolicyExpr tableId pd)
+      | pd <- Set.toList desiredPolicies
+      ]
```
telser:
```suggestion
    addPolicySteps = fmap mkStep $ Set.toList desiredPolicies
    mkStep pd = mkMigrationStepWithType AddPolicyDefinitions (Schema.mkCreatePolicyExpr tableId pd)
```

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1749,128 @@ findCurrentNamespace = do
...
+        let
+          nonEmptyTableNames = firstName :| restNames
+          schemaNameStr = PgCatalog.namespaceNameToString currentNamespace
+          whereCondition =
+            Orville.fieldIn pgPolicyTableNameField nonEmptyTableNames
+              `Orville.andExpr` Orville.fieldEquals pgPolicySchemaNameField (T.pack schemaNameStr)
+
+        pgPolicies <- Orville.findEntitiesBy pgPoliciesTable (Orville.where_ whereCondition)
+        pure $
+          foldl
```
telser: Is there a reason to not use foldl' here?

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1749,128 @@ findCurrentNamespace = do
...
+    roles = Set.fromList . fmap T.unpack $ rolesList
+
+    -- If the only role is "public", this means no specific roles were set
+    mbRoles =
+      if roles == Set.singleton "public"
+        then Nothing
+        else Just roles
+
+    mbUsing =
+      fmap (RawSql.unsafeSqlExpression . (\s -> "(" <> s <> ")") . T.unpack) $ pgPolicyQual row
```
telser: Why not use the parenthesized helper we have for this?

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1749,128 @@ findCurrentNamespace = do
         throwIO $ UnableToDiscoverCurrentSchema "No results returned by current_schema query"
       _ ->
         throwIO $ UnableToDiscoverCurrentSchema "Multiple results returned by current_schema query"
+
+-- Policy migration support
```
telser, line 1802: I think we're past time to break this module up. I don't think we need to make it part of this PR but we definitely should do it soon.

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1749,128 @@ findCurrentNamespace = do
...
+  PgPolicyRow
+    <$> Orville.marshallReadOnlyField pgPolicySchemaNameField
+    <*> Orville.marshallReadOnlyField pgPolicyTableNameField
+    <*> Orville.marshallReadOnlyField pgPolicyPolicyNameField
+    <*> Orville.marshallReadOnlyField pgPolicyRolesField
+    <*> Orville.marshallReadOnlyField pgPolicyQualField
+    <*> Orville.marshallReadOnlyField pgPolicyWithCheckField
+
+pgPolicySchemaNameField :: Orville.FieldDefinition Orville.NotNull T.Text
+pgPolicySchemaNameField = Orville.unboundedTextField "schemaname"
```
telser: I think we probably should have a 'name' field helper for things that are of that special type if only to provide signal to ourselves where we have them because they aren't the same as 'text'. By default 'name' is 64 bytes instead of unbounded with the 64 adjustable at sever compile time. See [docs](https://www.postgresql.org/docs/current/datatype-character.html) if anyone is curious. Using this read only as above obviously won't hurt us, just as a reminder to ourselves that we could model this a bit more precisely.

orville-postgresql/src/Orville/PostgreSQL/Expr/TableDefinition.hs
```diff
@@ -458,3 +460,27 @@ truncateTablesExpr :: NonEmpty (QualifiedOrUnqualified TableName) -> TruncateTab
...
+  specified table.
+
+  @since 1.2.0.0
+-}
+enableRowLevelSecurityExpr :: QualifiedOrUnqualified TableName -> AlterTableExpr
+enableRowLevelSecurityExpr tableName =
+  AlterTableExpr $
+    RawSql.fromString "ALTER TABLE "
+      <> RawSql.toRawSql tableName
+      <> RawSql.fromString " ENABLE ROW LEVEL SECURITY"
```
telser:
```suggestion
  alterTableExpr tableName . pure . AlterTableAction $ RawSql.fromString " ENABLE ROW LEVEL SECURITY"
```

orville-postgresql/src/Orville/PostgreSQL/Expr/TableDefinition.hs
```diff
@@ -458,3 +460,27 @@ truncateTablesExpr :: NonEmpty (QualifiedOrUnqualified TableName) -> TruncateTab
...
+  specified table.
+
+  @since 1.2.0.0
+-}
+disableRowLevelSecurityExpr :: QualifiedOrUnqualified TableName -> AlterTableExpr
+disableRowLevelSecurityExpr tableName =
+  AlterTableExpr $
+    RawSql.fromString "ALTER TABLE "
+      <> RawSql.toRawSql tableName
+      <> RawSql.fromString " DISABLE ROW LEVEL SECURITY"
```
telser:
```suggestion
  alterTableExpr tableName . pure . AlterTableAction $ RawSql.fromString " DISABLE ROW LEVEL SECURITY"
```

orville-postgresql/src/Orville/PostgreSQL/Expr/PolicyExpr/CreatePolicyExpr.hs
```diff
@@ -0,0 +1,237 @@
...
+newtype PolicyUsingExpr
+  = PolicyUsingExpr RawSql.RawSql
+  deriving
+    ( -- | @since 1.2.0.0
+      RawSql.SqlExpression
+    )
+
+-- | @since 1.2.0.0
+instance Eq PolicyUsingExpr where
+  (==) = on (==) (BS.map toLowerWord8 . RawSql.toExampleBytes)
```
telser: To my knowledge we haven't put Eq/Ord on our Expr types thus far. This could cause some issues if we aren't now very careful about 'toExampleBytes' which I don't know that we have been so far. Not saying we _have_ to change here, but this pattern is potentially dangerous. @qxjit might have thoughts.

orville-postgresql/src/Orville/PostgreSQL/Expr/PolicyExpr/CreatePolicyExpr.hs
```diff
@@ -0,0 +1,237 @@
...
+@since 1.2.0.0
+-}
+createPolicyExpr ::
+  PolicyName ->
+  QualifiedOrUnqualified TableName ->
+  Maybe [PolicyRole] ->
+  Maybe PolicyUsingExpr ->
+  Maybe PolicyCheckExpr ->
+  CreatePolicyExpr
+createPolicyExpr name tableName mbRoles mbUsing mbCheck =
```
telser: I think we should go ahead and have the PERMISSIVE/RESTRICTIVE support at this layer, even if we don't expose it, yet, at the higher level. That way we have the lowest level fully fleshed out and won't have a breaking change if/when we add support at the higher levels.

telser (changes requested):
A smattering of things of various importance. Really looking forward to get this in.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
```diff
@@ -642,3 +658,93 @@ collectSqlValue entry encodeRest entity =
...
+  TableDefinition key writeEntity readEntity ->
+  TableDefinition key writeEntity readEntity
+addTablePolicies policyDefs tableDef =
+  let
+    addPolicy policyDef =
+      Map.insert (policyDefinitionPolicyName policyDef) policyDef
+  in
+    tableDef
+      { i_tablePolicyDefinitions = foldr addPolicy (i_tablePolicyDefinitions tableDef) policyDefs
+      }
```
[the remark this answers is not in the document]
onslaughtq: @qxjit @telser Is copilot correct here? If so, then we have this same issue on functions, such as `addTableIndexes`, that have already been shipped as part of 1.0. If Copilot's correct and this behaviour is intentional, than maybe we just need to up the documentation?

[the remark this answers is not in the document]
telser: I'm pretty sure copilot is correct here and we have two options I think, either use `foldl'` in both places or update the docs in both places. I think I would prefer to use `foldl'` and keep the docs as they are because, as copilot notes, it is more in line with the usual last one wins.

Also, I'll note foldr probably would be better if we were dealing with an infinite list, but if someone tries to have an infinite list of policy definitions then us handling it poorly is the least of their worries.

orville-postgresql/orville-postgresql.cabal
```diff
@@ -146,13 +146,18 @@ library
...
       Orville.PostgreSQL.Expr.Internal.Name.IndexName
+      Orville.PostgreSQL.Expr.Internal.Name.PolicyName
       Orville.PostgreSQL.Expr.Internal.Name.Qualified
       Orville.PostgreSQL.Expr.Internal.Name.SavepointName
       Orville.PostgreSQL.Expr.Internal.Name.SchemaName
       Orville.PostgreSQL.Expr.Internal.Name.SequenceName
       Orville.PostgreSQL.Expr.Internal.Name.TableName
       Orville.PostgreSQL.Expr.Internal.Name.TriggerName
       Orville.PostgreSQL.Expr.Internal.Name.WindowName
+      Orville.PostgreSQL.Expr.PolicyExpr
```
telser: I think you meant for at least this module to be in the exposed module list. Per the comment above about public/private docs it seems like we might want the others to _not_ be in the exposed list though.

It is on my list to do a pass over this list overall before we cut another release, so I don't think we should worry about anything else in the list for this pr.

Oh and worth noting that to change this would be changing in the package.yaml file.

telser (changes requested):
We should figure out the module exposure we want, but my previous concern is gone I think.

orville-postgresql/src/Orville/PostgreSQL/Schema/TableDefinition.hs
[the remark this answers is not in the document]
onslaughtq: Fixed to use foldl'

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
```diff
@@ -1673,3 +1798,148 @@ findCurrentNamespace = do
...
+        else Schema.PolicyPermissive
+
+    -- pg_policies.cmd contains 'ALL', 'SELECT', 'INSERT', 'UPDATE' or 'DELETE'
+    command =
+      case T.unpack (T.toUpper (PgCatalog.pgPolicyCmd row)) of
+        "SELECT" -> Schema.PolicyCommandSelect
+        "INSERT" -> Schema.PolicyCommandInsert
+        "UPDATE" -> Schema.PolicyCommandUpdate
+        "DELETE" -> Schema.PolicyCommandDelete
+        _ -> Schema.PolicyCommandAll
```
qxjit: Is it worth making this an error if it's anything other than `"ALL"`?

orville-postgresql/test/Test/AutoMigration.hs
```diff
@@ -1453,6 +1478,647 @@ prop_dropsUnrequestedTriggers =
...
+              mkNamedPolicy "policy_two" (Just Orville.PolicyRestrictive)
+            , mkNamedPolicy "policy_two" Nothing
+            ]
+            policyTestTable
+      policies = Orville.tablePolicies tableDef
+
+    Map.keys policies === ["policy_one", "policy_two"]
+    fmap Orville.policyDefinitionPermission (Map.lookup "policy_one" policies) === Just Orville.PolicyRestrictive
+    fmap Orville.policyDefinitionPermission (Map.lookup "policy_two" policies) === Just Orville.PolicyPermissive
+
```
qxjit, line 2121: It's probably time to split this up too, but that can also be done later.

qxjit (approved):
💯  ❤️

orville-postgresql/src/Orville/PostgreSQL/AutoMigration.hs
onslaughtq: If we make this an error, should we also error if the roletext is not "public" and if the permission is not "RESTRICTIVE" or "PERMISSIVE"?

qxjit: Probably. We should be consistent, whichever way we go.

